PR #1, opened 2016-10-06
Initial commit of Analyzer-based Linter.  Ported over first linter 'unbalanced-delimiters'

usergenic:
This change is [Reviewable](https://reviewable.io/reviews/polymer/polymer-linter/1)


src/linter.ts
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  /**
+   * Given an array of filenames, lint the files and return an array of all
+   * warnings produced evaluating the linter rules.
+   */
+  public async lint(files: string[]): Promise<Warning[]> {
+    let warnings: Warning[] = [];
+    for (const file of files) {
+      const document: Document = await this.analyzer.analyze(file);
```
rictic: To catch stuff like syntax errors that stop us from creating a Document object, something like this is needed:

``` typescript
let document: Document;
try {
  document = await this.analyzer.analyze(file);
} catch (error) {
  if (error instanceof WarningCarryingException) {
    warnings.push(error.warning);
    continue;
  }
  throw error;
}
```

src/linter.ts
```diff
@@ -0,0 +1,46 @@
...
+  public rules: Rule[];
+
+  constructor(rules: Rule[], analyzer?: Analyzer) {
+    this.analyzer = analyzer || new Analyzer({urlLoader: new FSUrlLoader()});
+    this.rules = Array.from(rules);
+  }
+
+  /**
+   * Given an array of filenames, lint the files and return an array of all
+   * warnings produced evaluating the linter rules.
```
rictic, line 36: Might be worth noting that the files need to be ones that the `analyzer` instance can resolve?

src/rule.ts
```diff
@@ -0,0 +1,22 @@
...
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+'use strict';
+
+import {Document} from 'polymer-analyzer/lib/model/document';
+import {Warning} from 'polymer-analyzer/lib/warning/warning';
+
+// TODO(usergenic): Linter rules should have a standard interface for
+// configuring Warning severity and for setting their options.
```
rictic: I suspect we'll want to do warning severity filtering and modification in a central place, downstream of individual rules

src/rules/unbalanced-delimiters.ts
```diff
@@ -0,0 +1,132 @@
...
+    const parsedHtml = document.parsedDocument;
+
+    // This linter is only relevant to html documents, so exit out unless the
+    // provided document is one.
+    if (!(parsedHtml instanceof ParsedHtmlDocument)) {
+      return warnings;
+    }
+
+    // Address binding expressions only for polymer elements.
+    for (const element of document.getByKind('polymer-element')) {
```
rictic: Would it be possible to structure this rule with a `getDataBindingExpressions` method? That'd be really broadly useful.

Hm, I guess not, as we're looking for stuff that's only _almost_ legal data binding expressions.

src/rules/unbalanced-delimiters.ts
```diff
@@ -0,0 +1,132 @@
...
+      console.assert(element instanceof PolymerElement);
+
+      for (const warning of this._warningsForPolymerElement(
+               parsedHtml, element)) {
+        warnings.push(warning);
+      }
+    }
```
rictic, line 49: We should also check inside dom-bind templates.

src/rules/unbalanced-delimiters.ts
```diff
@@ -0,0 +1,132 @@
...
+  private _warningsForPolymerElement(
+      parsedHtml: ParsedHtmlDocument,
+      element: PolymerElement): Warning[] {
+    let warnings: Warning[] = [];
+
+    if (!element.domModule) {
+      return warnings;
+    }
+
+    const template = dom5.query(element.domModule, matchers.isTemplate);
```
rictic: An element's dom-module could be in another file. I'd recommend doing a  `getByKind('dom-module')` instead.

src/rules/unbalanced-delimiters.ts
```diff
@@ -0,0 +1,132 @@
...
+    if (!template) {
+      return warnings;
+    }
+
+    const templateContent =
+        parse5.treeAdapters.default.getTemplateContent(template);
+
+    dom5.nodeWalkAll(templateContent, (node: parse5.ASTNode) => {
```
rictic: I think we want to recurse into inner template contents as well. (or maybe just `dom-if` and `dom-repeat`? but with that changing in 2.0, probably easier to just recurse into all template elements)

src/rules/unbalanced-delimiters.ts
```diff
@@ -0,0 +1,132 @@
...
+          sourceRange:
+              parsedHtml.sourceRangeForAttribute(element, attr.name)!
+        });
+      }
+    }
+    return warnings;
+  }
+
+  private _extractBadBindingExpression(text: string): string|undefined {
+    // 4 cases, {{}, {}}, [[], []]
```
rictic, line 118: If there are just four cases, could we check for them individually? It would make for more legible code (and I say this as a generally regex-loving weirdo), and we could provide better warning messages. For me, I think the ideal message here would be:

`Invalid polymer expression delimiters. You put '{{}' are you missing a closing }`

I think it's worthwhile to say that they're polymer expression delimiters (and not like, a missing closing paren _inside_ the polymer expression), and to say what we saw and how to fix.

src/rules/unbalanced-delimiters.ts
```diff
@@ -0,0 +1,132 @@
...
+      parsedHtml: ParsedHtmlDocument,
+      element: parse5.ASTNode): Warning[] {
+    const warnings: Warning[] = [];
+    for (const attr of element.attrs) {
+      if (this._extractBadBindingExpression(attr.value || '')) {
+        warnings.push({
+          code: 'unbalanced-delimiters',
+          message: `Expression ${attr
+                       .value
+                   } has unbalanced delimiters in attribute ${attr.name}`,
```
rictic: Generally I'd leave out information in the message that's expressed well in the sourceRange

rictic (changes requested):

src/linter.ts
usergenic: Ok.  I saw that in the linter demo code-- I'll yoink it for here.

src/rules/unbalanced-delimiters.ts
usergenic: I mostly copied and pasted this bit from polylint and hadn't really even thought of improving the matcher given there were 13 linters left to port over, but you make a good point about maintainability + legibility + much better result.

package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "polymer-linter",
+  "version": "1.0.0",
```
justinfagnani: Let's start with a pre-release tag: 1.0.0-alpha.1

src/linter.ts
```diff
@@ -0,0 +1,46 @@
...
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+'use strict';
```
justinfagnani: not necessary, tsc adds this for us.

src/linter.ts
justinfagnani: Hopefully this is temporary?

src/rules/unbalanced-delimiters.ts
```diff
@@ -0,0 +1,132 @@
...
+    if (!(parsedHtml instanceof ParsedHtmlDocument)) {
+      return warnings;
+    }
+
+    // Address binding expressions only for polymer elements.
+    for (const element of document.getByKind('polymer-element')) {
+      console.assert(element instanceof PolymerElement);
+
+      for (const warning of this._warningsForPolymerElement(
+               parsedHtml, element)) {
```
justinfagnani: really funky wrapping here, can you add a variable to hold the warnings before the loop?

src/rules/unbalanced-delimiters.ts
justinfagnani: in theory one can vend other directive-style templates, though no one really does. So traversing into all templates is the easy and slightly more correct bet.

src/linter.ts
usergenic: Done.

justinfagnani (commented):
Man, I really have to get better about submitting my review comments!

src/linter.ts
rictic: What's the alternative? We've asked the analyzer to analyze a file, but it has a syntax error.

I guess we could return some sort of mostly empty Document object that has the parse error as a warning on it, but for most use cases I think that's a foot gun.

src/rule.ts
usergenic: That may be true.  Rule configurability isn't an issue yet, just something I anticipate.  Having a standard interface in constructor will be useful to configure Rule where they have configurability, like keyword blacklists or whatever.  This also would allow polymer-linter.json or poylmer.json linter stanza to include per-rule config bits that are handed to the constructors in an obvious way.

Like, maybe instantiating the linter by way of a linter config would be similar to the way we do properties in Polymer.  A rule class name on the left and in the typical case, the value on the right is just `true` or `false` and in the configurable case its an object i.e. `{ whitelist: ['h1','p','em','strong'] }`

Anyways that's why I made them classes with a standard `check` method.

src/rules/unbalanced-delimiters.ts
usergenic: It's a tricky one.  If we extend analyzer to include binding expressions as features, it seems like it maybe shouldn't return binding expressions which are not properly formed, since that would be confusing.  It totally depends on the way Analyzer treats returning features from scanners that are improperly formed or invalid in some way as a matter of convention.

Do you have any thoughts on whether Analyzer would report bad binding expressions as features?

src/rules/unbalanced-delimiters.ts
usergenic: Great point- I totally forgot dom-bind was a thing.  Adding.

src/rules/unbalanced-delimiters.ts
usergenic: Okay.  Converting.

src/rules/unbalanced-delimiters.ts
usergenic: That's a good rule to live by with these.  Unredundantizing.

src/rules/unbalanced-delimiters.ts
rictic: Chatted out in person. I think consensus was to add this logic into the analyzer at some point, but it's fine to land it here for now.

package.json
usergenic: Done.

src/rule.ts
rictic: Definite +1 to making them configurable classes. I think we're going to want to add `code` and `description` properties to the Rule classes too.

src/linter.ts
usergenic: Oh!  Good to know.  Removed from all the things.

src/linter.ts
usergenic: I have a catch now for the specific `Error` subclass `WarningCarryingException` as in the demo linter file in analyzer.

src/rules/unbalanced-delimiters.ts
usergenic: Take it up with clang-format!  But okay :)

src/rules/unbalanced-delimiters.ts
usergenic: Since I also had to support `dom-bind` templates, I dropped the polymer element stuff and just search for `<template>` inside `dom-modules` and `<template is="dom-bind">` without using the analyzed feature for now.  Will migrate to feature when `dom-bind` templates are treated as such.

src/rules/unbalanced-delimiters.ts
usergenic: Recursing without discriminating.  Added a comment to point out there's no condition applied and one may be considered in the future.

src/expressions.ts
```diff
@@ -0,0 +1,168 @@
...
+  value?: string|number;
+  literal?: boolean;
+  structured?: boolean;
+  wildcard?: boolean;
+}
+
+function primaryName(expression: string): string {
+  // TODO(usergenic): Remove this commented out section copied in from polylint
+  // if (expression.name) {
+  //   expression = expression.name;
```
rictic, line 49: Looks like this was sometimes an `estree.Identifier`, I say aggressively delete.

src/expressions.ts
```diff
@@ -0,0 +1,168 @@
...
+// https://github.com/Polymer/polymer-analyzer/issues/351
+
+/**
+ * A parsed polymer binding expression
+ * @param {Array.<string>} keys The keys referenced by this expression.
+ * @param {Array.<string>} methods The methods referenced by this expression.
+ * @param {string} type One of 'computed', 'literal', or 'reference'
+ * @param {string} raw  The unparsed expression
+ */
+export class ParsedExpression {
```
rictic, line 25: Recommend instead making this:

`type ParsedExpression = Computed | Literal | Reference`

And make each of those three types an interface with a string literal `type` field. That way we can separate out the fields and type safety will help us in a lot of ways.

This is how we're typing estree, and it's paid off hugely. It lets you switch over `expression.type`.. lots of stuff.

src/linter.ts
rictic: Double check that this change made it in.

src/rules/native-attribute-binding.ts
```diff
@@ -0,0 +1,140 @@
...
+               (e) => true)) {
+        for (const attr of element.attrs) {
+          if (this._isBindingExpression(attr.value) &&
+              !this._canAttributeUseNativeBinding(element, attr)) {
+            const parsedHtml = document.parsedDocument;
+            if (!(parsedHtml instanceof ParsedHtmlDocument)) {
+              continue;
+            }
+            warnings.push({
+              code: 'native-attribute-binding',
```
rictic, line 106: Here and elsewhere in this file, this isn't about `native` attribute binding so much as attributes which have no corresponding property.

Recommend updating the names, code, and message here.

Also, do we need to add the attr value and name here? With the new warning printer it will show you the line of code with full info.

src/rules/native-attribute-binding.ts
```diff
@@ -0,0 +1,140 @@
...
+      element: parse5.ASTNode,
+      attr: parse5.ASTAttribute): boolean {
+    const name: string = attr.name.toLowerCase();
+    if (name === 'for' && dom5.predicates.hasTagName('label')(element)) {
+      return false;
+    }
+    if (nativeAttributes.has(name)) {
+      return false;
+    }
+    if (name.indexOf('data-') === 0 && attr.name[name.length - 1] !== '$') {
```
rictic, line 130: Why are we checking whether attr.name ends with `$` here? Shouldn't it be checked elsewhere already?

src/rules/native-attribute-binding.ts
```diff
@@ -0,0 +1,140 @@
...
+      return false;
+    }
+    return true;
+  }
+
+  private _isBindingExpression(expression: string): boolean {
+    const bindingMatch = expressionParser.extractBindingExpression(expression);
+    return !!bindingMatch || false;
```
rictic, line 138: AFAIK `!!foo || false` is equal to `!!foo` for all values of `foo`.

src/rules/unbalanced-delimiters.ts
usergenic: Step 1 complete: I made the messages nicer.   I have not made bad binding expression match code nicer yet.

test/sample/my-element-collection.html
```diff
@@ -0,0 +1,17 @@
+<link rel="import" href="../../polymer/polymer.html">
```
rictic, line 1: Is this used in a later commit?

src/test/native-attribute-binding_test.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    assert.equal(warnings.length, 1);
+
+    assert.deepEqual(warnings[0], {
+      code: 'native-attribute-binding',
+      message:
+          'The expression [[myVars]] bound to attribute \'class\' should use $= instead of =',
+      severity: Severity.ERROR,
+      sourceRange: {
+        end: {column: 28, line: 11},
```
rictic, line 36: Recommend using the WarningPrinter#printUnderlinedText() trick I used in my analyzer PR recently to test these source ranges. Makes them more maintainable and more easily reviewed.

test/sample/imports/bind-to-data.html
```diff
@@ -0,0 +1,30 @@
...
+@license
+Copyright (c) 2015 The Polymer Project Authors. All rights reserved.
+This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+Code distributed by Google as part of the polymer project is also
+subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+-->
+<dom-module id="bind-to-data">
+  <template>
```
rictic, line 11: Consider putting this into a dom-bind instead

test/sample/jsconf/README.md
```diff
@@ -0,0 +1,12 @@
+# JSConformance sample files
```
rictic, line 1: Wow, was this actually implemented in original polylint?

rictic (changes requested):

rictic:
This is good work, but would it be possible to break it up into separate PRs? You're adding code at a faster rate than we can get it reviewed, which is a temporary situation, but it also means that nothing can land until we can catch up with your full speed run!

rictic:
(I resolved the conflicts with master through the github UI, but it looks like it created a merge commit. Might want to rebase that out.)

rictic:
Also be quiet googlebot.

usergenic:
closing this PR because @rictic's work obviates this bit-o scaffolding

PR #3, opened 2016-10-17
Undefined elements rule

43081j:
Added an undefined elements rule for finding references to elements which don't exist.

Note that this requires the latest polymer-analyzer (master, not yet tagged) so won't work with the current version in our package.json.

This makes use of the new element reference scanner added in polymer-analyzer.

---
This change is [Reviewable](https://reviewable.io/reviews/polymer/polymer-linter/3)


rictic:
Ok, the project is now ready to receive this contribution! Mind rebasing this off master?

I think the changes needed are minor. Just need to add a description of the rule, call registry.register() on it, and add it to the collections in collections.ts, then this should be good to go!

43081j:
@rictic here you go. rebased off master, shuffled the code around to match the new structure + changed it to be a `HtmlRule` instead of just a `Rule`.

Also formatted the code in it.

src/html/undefined-elements.ts
```diff
@@ -0,0 +1,61 @@
...
+    const warnings: Warning[] = [];
+
+    const elements =
+        Array.from(document.getByKind('element')).map((e) => e.tagName);
+
+    const refs = document.getByKind('element-reference');
+
+    for (const ref of refs) {
+      const elementRef = ref as ElementReference;
+      if (elements.indexOf(elementRef.tagName) === -1) {
```
rictic: Recommend instead:

`document.getById('element', elementRef.tagName, {imported: true, externalPackages: true}).size > 0`

That call is very fast, and will find elements anywhere that's reachable from the document.

src/html/undefined-elements.ts
```diff
@@ -0,0 +1,61 @@
...
+import {registry} from '../registry';
+
+import {HtmlRule} from './rule';
+
+import stripIndent = require('strip-indent');
+
+export class UndefinedElements extends HtmlRule {
+  code = 'undefined-elements';
+  description = stripIndent(`
+      Warns for elements which have no definition.
```
rictic: We should mention that it looks at element references from html. Also it's good to call trim() after stripIndent. How about:

```typescript
description = stripIndent(`
  Warns when an HTML tag refers to a custom element with no known definition.
`).trim()
```

src/html/undefined-elements.ts
```diff
@@ -0,0 +1,61 @@
...
+      Promise<Warning[]> {
+    const warnings: Warning[] = [];
+
+    const elements =
+        Array.from(document.getByKind('element')).map((e) => e.tagName);
+
+    const refs = document.getByKind('element-reference');
+
+    for (const ref of refs) {
+      const elementRef = ref as ElementReference;
```
rictic: I think you can drop this line, type inference should have your back now.

43081j:
@rictic this is sorted now, suggestions have been added!

rictic:
Cool, tests look good, and the integration tests are clean too. Great PR!

PR #10, opened 2016-12-21
Initial scaffolding for the new linter, without any rules implemented yet

rictic:
Extracted by rictic from usergenic's https://github.com/Polymer/polymer-linter/pull/1 PR

### DIFF
--- a/src/linter.ts
+++ b/src/linter.ts
@@ -1,0 +1,57 @@
+/**
+ * @license
+ * Copyright (c) 2014 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at
+ * http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+import {Analyzer} from 'polymer-analyzer';
+import {Document} from 'polymer-analyzer/lib/model/document';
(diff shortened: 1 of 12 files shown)


src/linter.ts
```diff
@@ -0,0 +1,57 @@
...
+    let warnings: Warning[] = [];
+    for (const file of files) {
+      let document: Document;
+      try {
+        document = await this.analyzer.analyze(file);
+        for (const rule of this.rules) {
+          warnings = warnings.concat(await rule.check(document));
+        }
+      } catch (error) {
+        if (error instanceof WarningCarryingException) {
```
usergenic: Can we catch per-rule?  That way all rules will try to run if one fails.

usergenic (changes requested):

src/linter.ts
usergenic: If not here, lets file issue and move that code in separate PR?

rictic: +1, reworked how we do error handling here, PTAL

src/linter.ts
```diff
@@ -0,0 +1,76 @@
...
+   */
+  public async lint(files: string[]): Promise<Warning[]> {
+    let warnings: Warning[] = [];
+    const analysisResult = await this._analyzeAll(files);
+    const documents = analysisResult.documents;
+    const analysisWarnings = analysisResult.warnings;
+    warnings = warnings.concat(analysisWarnings);
+    for (const document of documents) {
+      for (const rule of this._rules) {
+        warnings = warnings.concat(await rule.check(document));
```
usergenic: Can we add a TODO (or implement?) here to catch exceptions for the `rule.check()` call?  I'd be nice if we had a standard warning to push like we do in the `_analyzeAll` call.

rictic: +1, done

usergenic (approved):
This is all sorts of LGTM

rictic:
Thanks for the review!

PR #11, opened 2016-12-22
Add some tests for Linter by itself.

rictic:
Also set up test scaffolding.

- [x] CHANGELOG.md has not been updated. Internal change, no user-visible changes.

---
This change is [Reviewable](https://reviewable.io/reviews/polymer/polymer-linter/11)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 lib
 node_modules
+yarn.lock
--- a/src/test/linter_test.ts
+++ b/src/test/linter_test.ts
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright (c) 2014 The Polymer Project Authors. All rights reserved.
+ * Copyright (c) 2017 The Polymer Project Authors. All rights reserved.
  * This code may only be used under the BSD style license found at
  * http://polymer.github.io/LICENSE.txt
  * The complete set of authors may be found at
@@ -23,13 +23,16 @@ import {Linter} from '../linter';
 const fixtures_dir = path.join(__dirname, '..', '..', 'test');
 
 suite('Linter', () => {
+
   suite('.lint', () => {
     let analyzer: Analyzer;
     let warningPrinter: WarningPrinter;
+
     setup(() => {
       analyzer = new Analyzer({urlLoader: new FSUrlLoader(fixtures_dir)});
       warningPrinter = new WarningPrinter(null as any, {analyzer: analyzer});
     });
+
     test('works in the trivial case', async() => {
       const linter = new Linter([], analyzer);
       const warnings = await linter.lint([]);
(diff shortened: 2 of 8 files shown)


rictic:
Activated travis. Bumping to see if it notices.

.gitignore
```diff
@@ -1,2 +1,3 @@
 lib
 node_modules
+yarn.lock
```
abdonrd, line 3: Why ignore `yarn.lock`?
https://yarnpkg.com/blog/2016/11/24/lockfiles-for-all

rictic: I don't think that this is as open and shut as that blog post makes it out to be.

When there's a choice between taking pain upon ourselves vs pushing pain onto our users, I think the ratio has to be pretty large for it to be net positive for us to push the pain down onto users.

Having travis notify us about upstream changes that break us is pretty nice as well, as it happens right when someone is actually working on the repo, and a failing test is probably going to be easier to debug than an issue as reported by a user.

src/test/linter_test.ts
```diff
@@ -0,0 +1,65 @@
+/**
+ * @license
+ * Copyright (c) 2014 The Polymer Project Authors. All rights reserved.
```
justinfagnani: 2017

src/test/linter_test.ts
```diff
@@ -0,0 +1,65 @@
...
+
+import {assert} from 'chai';
+import * as path from 'path';
+import {Analyzer} from 'polymer-analyzer';
+import {FSUrlLoader} from 'polymer-analyzer/lib/url-loader/fs-url-loader';
+import {WarningPrinter} from 'polymer-analyzer/lib/warning/warning-printer';
+
+import {Linter} from '../linter';
+
+const fixtures_dir = path.join(__dirname, '..', '..', 'test');
```
justinfagnani, line 23: `path.resolve(__dirname, '../../test')`

src/test/linter_test.ts
```diff
@@ -0,0 +1,65 @@
...
+
+suite('Linter', () => {
+  suite('.lint', () => {
+    let analyzer: Analyzer;
+    let warningPrinter: WarningPrinter;
+    setup(() => {
```
justinfagnani, line 31: lets add some newlines and let this section breathe a little

justinfagnani (approved):

src/test/linter_test.ts
rictic: Will that work on Windows?

src/test/linter_test.ts
rictic: done

src/test/linter_test.ts
rictic: Done

PR #18, opened 2017-02-09
Add rule collections and a registry.

rictic:

---
This change is [Reviewable](https://reviewable.io/reviews/polymer/polymer-linter/18)

### DIFF
--- a/src/registry.ts
+++ b/src/registry.ts
@@ -16,18 +16,16 @@ import {Rule, RuleCollection} from './rule';
 import {stripWhitespace} from './util';
 
 /**
- * A singleton class where lint rules and rule collections can register
+ * A centralized place where lint rules and rule collections can register
  * themselves and you can get a collection of rules by querying.
+ *
+ * Almost all users should not construct their own registry, but instead use
+ * the instance exported as `registry` from this module.
  */
 export class LintRegistry {
-  private static readonly _singleton = new LintRegistry();
-  static get instance() {
-    return LintRegistry._singleton;
-  }
-
   private _all = new Map<string, Rule|RuleCollection>();
 
-  private constructor() {
+  constructor() {
   }
 
   /**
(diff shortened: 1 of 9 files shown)


src/registry.ts
```diff
@@ -0,0 +1,89 @@
...
+
+import {Rule, RuleCollection} from './rule';
+import {stripWhitespace} from './util';
+
+/**
+ * A singleton class where lint rules and rule collections can register
+ * themselves and you can get a collection of rules by querying.
+ */
+export class LintRegistry {
+  private static readonly _singleton = new LintRegistry();
```
justinfagnani: I don't think you really need to both expose a registry variable and try to make the class a singleton in it's implementation. Exporting the `registry` binding is enough. It's similar to `CustomElementRegistry`, there's a built-in global singleton, but you could create other instances if you wanted.

src/registry.ts
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  private _getRules(ruleCodes: string[], alreadyExpanded: Set<string>):
+      Set<Rule> {
+    ruleCodes = ruleCodes.filter(p => !alreadyExpanded.has(p));
+
+    const results = new Set<Rule>();
+
+    for (const code of ruleCodes) {
+      for (const rule of this._getRulesForCode(code, alreadyExpanded)) {
```
justinfagnani: pass results down here

src/registry.ts
```diff
@@ -0,0 +1,89 @@
...
+        results.add(rule);
+      }
+    }
+
+    return results;
+  }
+
+  private _getRulesForCode(code: string, alreadyExpanded: Set<string>) {
+    alreadyExpanded.add(code);
+    const ruleOrCollection = this._all.get(code);
```
justinfagnani: given then corecursion here, and the relative simplicity of this method, it might be easier to follow if this were just inlined into `_getRules`

src/rule.ts
```diff
@@ -35,3 +35,36 @@ export abstract class Rule {
...
+  /**
+   * A unique string identifying this collection. Uses the same namespace as
+   * Rules.
+   */
+  code: string;
+
+  /**
+   * Describes the rule collection.
+   *
+   * Who should use it? When? What should they expect?
```
justinfagnani: "A description should say..."

justinfagnani (approved):

src/registry.ts
rictic: Fair. Done.

rictic: Fair, done.

src/registry.ts
rictic: done

src/registry.ts
rictic: fair. done.

src/rule.ts
rictic: revised

PR #20, opened 2017-02-10
Prepare for the v0.1.0 release

rictic:

---
This change is [Reviewable](https://reviewable.io/reviews/polymer/polymer-linter/20)


src/test/integration_test.ts
```diff
@@ -86,6 +86,35 @@ const fileSpecificIgnoredCodes: {[path: string]: Set<string>} = {
   // https://github.com/PolymerElements/iron-resizable-behavior/pull/25
   'iron-resizable-behavior/test/test-elements.html':
       new Set(['unknown-polymer-behavior']),
+
+  // https://github.com/Polymer/polymer-analyzer/issues/458
+  'hydrolysis/custom_typings/escodegen.d.ts': new Set(['unable-to-analyze']),
```
justinfagnani, line 91: Why are the integration tests running on .d.ts files?

rictic: analyzePackage() operates over .d.ts files if there are .d.ts files in the package.

Maybe it shouldn't.

justinfagnani: Is this because we added the typescript parser? `.d.ts` is an unfortunate extension... we need to filter it out from files that end in just `.ts`

rictic: Yeah, it's analyzing the typescript and running into module resolution errors. I think it's reasonable to ignore these errors for now and take another look later.

rictic:
Changed to v0.1.0, ptal

justinfagnani (approved):

PR #21, opened 2017-02-14
Add accessor methods to registry.

rictic:
For the purpose of displaying documentation on registered rules and collections in the cli.

---
This change is [Reviewable](https://reviewable.io/reviews/polymer/polymer-linter/21)

### DIFF
--- a/src/registry.ts
+++ b/src/registry.ts
@@ -75,6 +75,16 @@ export class LintRegistry {
       }
     }
   }
+
+  allRules(): Iterable<Rule> {
+    return Array.from(this._all.values())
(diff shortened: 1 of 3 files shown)


src/registry.ts
```diff
@@ -75,6 +75,16 @@ export class LintRegistry {
       }
     }
   }
+
+  allRules(): Iterable<Rule> {
```
justinfagnani: these could be getters

justinfagnani (approved):

src/registry.ts
rictic: +1, done

PR #23, opened 2017-02-15
Add unbalanced polymer databinding delimitors rule.

rictic:

Steals from @usergenic's PR #1

---
This change is [Reviewable](https://reviewable.io/reviews/polymer/polymer-linter/23)

### DIFF
--- a/src/polymer/matchers.ts
+++ b/src/polymer/matchers.ts
@@ -1,0 +1,34 @@
+/**
+ * @license
+ * Copyright (c) 2014 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at
+ * http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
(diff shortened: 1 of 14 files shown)


rictic:
Fixes #30

src/polymer/matchers.ts
```diff
@@ -0,0 +1,59 @@
...
+export const isDomBindTemplate =
+    p.AND(p.hasTagName('template'), p.hasAttrValue('is', 'dom-bind'));
+export const isDomModuleTemplate = p.AND(
+    p.parentMatches(p.hasTagName('dom-module')), p.hasTagName('template'));
+export const isTemplate = p.hasTagName('template');
+export const isTemplateDescendant = p.parentMatches(isTemplate);
+
+export interface Template extends parse5.ASTNode { content: parse5.ASTNode; }
+
+export function getNestedDataBindingTemplates(node: parse5.ASTNode) {
```
usergenic: I feel like this needs commenting.  Also, I feel like matchers.ts should be all predicate functions, but maybe that wouldn't make sense for this one, so not a deal-breaker.

usergenic (changes requested):

src/polymer/matchers.ts
rictic: It turns out that this can be written more correctly as a predicate (as top level dom-if and dom-repeat is allowed).

usergenic: So clean!  Awesome.

usergenic (approved):
LGTM

PR #37, opened 2017-02-24
add dom module after polymer rule

### LINKED ISSUE #26: dom-module must come before the script that uses it
Port from polylint. See domModuleAfterPolymer here: https://github.com/PolymerLabs/polylint/blob/master/src/linters.ts#L457

43081j:

Fixes #26, added a rule for checking if a dom-module has been defined after its associated polymer element definition.

---
This change is [Reviewable](https://reviewable.io/reviews/polymer/polymer-linter/37)


src/html/dom-module-after-polymer.ts
```diff
@@ -0,0 +1,79 @@
...
+    const warnings: Warning[] = [];
+    const elements = document.getByKind('polymer-element');
+    const moduleForId = (id: string) =>
+        p.AND(p.hasTagName('dom-module'), p.hasAttrValue('id', id));
+
+    for (const element of elements) {
+      const domModule = document.getOnlyAtId(
+          'dom-module',
+          element.tagName!,
+          {imported: true, externalPackages: true});
```
43081j: i have to do this because for some reason `element.domModule` doesn't seem to get set, any idea why this is? analyser just doesn't find it at the min?

src/html/dom-module-after-polymer.ts
```diff
@@ -0,0 +1,79 @@
...
+
+      if (domModule && element.scriptElement) {
+        const otherDomModule = dom5.nodeWalkPrior(
+            element.scriptElement, moduleForId(element.tagName!));
```
43081j: is it safe to assume `tagName` always exists here? (its optional in the model). i was expecting `element.is` but that seems to have gone since i was last around this code...

src/html/dom-module-after-polymer.ts
```diff
@@ -0,0 +1,79 @@
...
+    const moduleForId = (id: string) =>
+        p.AND(p.hasTagName('dom-module'), p.hasAttrValue('id', id));
+
+    for (const element of elements) {
+      const domModule = document.getOnlyAtId(
+          'dom-module',
+          element.tagName!,
+          {imported: true, externalPackages: true});
+
+      if (domModule && element.scriptElement) {
```
43081j: **this currently seems to never be true in tests**, because it seems `scriptElement` doesn't get set. if you run the tests, you see by debugging that it is just empty/non-existent. i assumed i need the script element to walk back from (like the old linter did).

43081j:
i added a few comments to the changes.

this currently doesn't pass tests because of `scriptElement` never being populated (see comments).

cc @rictic

rictic:
Yeah, there's a couple of bugs in the analyzer right now around this. One of them will be fixed in the next release: https://github.com/Polymer/polymer-analyzer/pull/480

@FredKSchott started work on this pass as well. The tack I would take would be:

* index all dom modules in the document by id
* find all inline scripts and script src tags
* for each script, find all polymer-elements declared in it
* for each element:
  * find the dom-module with that element's id if any, and warn if [comparePositionAndRange](https://github.com/Polymer/polymer-linter/blob/master/src/polymer/set-unknown-attribute.ts#L240) says that the dom-module comes after the element's script tag.

This avoids several tricky issues, as it'd be easy to write this lint in a way that'd be N^2 (not bad in typical cases, but very bad if someone analyzes a bundled app), and as you've noticed we're currently not doing a good idea of filling out script information on polymer elements :X

43081j:
Is it not as simple as comparing the ranges we already have access to? (of the element and its dom module, if one is found).

in fact, i shouldn't even need the script element, right? I shouldn't need to do a walk up the tree, i already have the source range of both the element and the module. that's not enough?

just want to be sure i understand before any changes as your suggestion would complicate it a little more.

rictic:
We need the script element because of script import case. i.e. this should fail: `<script src="./my-element.js"></script><dom-module id="my-element"></dom-module>`, but the element's declaration source range will not be comparable to the dom-module's source range.

rictic:
Thanks for the PR! I wanted to get a version of this out there so we can reach feature parity with polylint so we can properly deprecate it, so this landed as #50. Sorry about the duplicated effort!

PR #35, opened 2017-02-23
Add a warning for databinding to undeclared properties/attributes

rictic:
- [x] CHANGELOG.md has been updated

Fixes #25

---
This change is [Reviewable](https://reviewable.io/reviews/polymer/polymer-linter/35)

PR #39, opened 2017-02-25
add unknown event declarations rule

### LINKED ISSUE #34: Warn for unknown / invalid event handler declarations
i.e.

```
  <paper-button on-unknown-event="foo"></paper-button>
                ~~~~~~~~~~~~~~~~

  <paper-button on-raised-changed="{{noBracesAround}}"></paper-button>
                                   ~~
```

43081j:

Fixes #34.

Still need to add a check for invalid event handler values (braces in there). Maybe I can just do a regex test on `\w+`.

---
This change is [Reviewable](https://reviewable.io/reviews/polymer/polymer-linter/39)


43081j:
This fails build because of line endings like the others. I guess its an error on my end, me committing bad line endings? I'm not sure.

Also isn't quite complete as said above, need to add a check for invalid values.

src/polymer/unknown-event-handler-declarations.ts
```diff
@@ -0,0 +1,72 @@
...
+import {registry} from '../registry';
+
+export class UnknownEventHandlerDeclarations extends Rule {
+  code = 'unknown-event-handler-declarations';
+  description = stripIndent(`
+      Warns for unknown event handler declarations:
+
+          <paper-button on-unknown-event="foo">
+
+      Correct syntax:
```
rictic: No need for a correct syntax section for this one

test/unknown-event-handler-declarations/unknown-event-handler-declarations.html
```diff
@@ -0,0 +1,33 @@
...
+The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+Code distributed by Google as part of the polymer project is also
+subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+-->
+
+<script>
+  Polymer({
+    is: 'unknown-event-handler-declarations'
+    /**
+      @event testevent
```
rictic: Oh nice! I didn't know we had `@event` hooked up in the analyzer already.

src/polymer/unknown-event-handler-declarations.ts
```diff
@@ -0,0 +1,72 @@
...
+        const events = definition.events.map((e) => e.name);
+
+        for (const attr of element.attributes) {
+          const eventName = attr.name.substr(3);
+          if (attr.name.indexOf('on-') === 0 &&
+              events.indexOf(eventName) === -1) {
+            warnings.push({
+              code: this.code,
+              message: stripWhitespace(`
+                  Unknown event handler "${attr.name}" on tag`),
```
rictic: I think we want something like:

```
`Tag ${element.tagName} is not known to emit an event named ${eventName}`
```

src/polymer/unknown-event-handler-declarations.ts
```diff
@@ -0,0 +1,72 @@
...
+              severity: Severity.WARNING,
+              sourceRange: attr.sourceRange
```
rictic: Is the source range for the attribute name available? If not, we could get it by doing something like `parsedDoc.getSourceRangeForAttributeName(attr.astNode, attr.name)`

src/polymer/unknown-event-handler-declarations.ts
```diff
@@ -0,0 +1,72 @@
...
+    const elements = document.getByKind('element-reference');
+
+    for (const element of elements) {
+      const definition = document.getOnlyAtId(
+          'polymer-element',
+          element.tagName,
+          {imported: true, externalPackages: true});
+
+      if (definition) {
+        const events = definition.events.map((e) => e.name);
```
rictic: Looks like we need to add built in events, as well as the events like `tap` that Polymer adds. Otherwise we'll see a lot of spurious warnings. See the [integration test results](https://travis-ci.org/Polymer/polymer-linter/jobs/205167886) for some examples. I think we can get a list of built-in html events from the html spec, like I did for attributes and properties here: https://github.com/Polymer/polymer-linter/pull/35/files#diff-84e513ab371344ae4799bcc2442486e6R28

rictic:
I've finished the flurry of mergings, can rebase in safety now :)

rictic:
Mind rebasing on top of #41? Between some changes to the analyzer and some new warnings from undefined elements (I thought there weren't any?) the integration test results are hard to read. They're not blocking submission, but it's useful to get a sense of whether we've got any false positives.

One thing that I can tell from just skimming the results though, we're going to want to add a set of shared events that are inherited from HTMLElement, like focus, as well as the events like tap that come in from Polymer's gestures support.

43081j:
yup i'll have a think about it today. we may just need an array of them all.

i'll give it a go

43081j:
@rictic will these do:

https://www.w3.org/TR/2013/WD-DOM-Level-3-Events-20131105/#h4_event-types-list

43081j:
after having run the integration tests, i've noticed also that we have another problem... bubbling events.

for example:

https://github.com/PolymerElements/paper-dropdown-menu/blob/master/paper-dropdown-menu-light.html#L286

`paper-menu-button` doesn't implement iron-selectable, there is no relation between `iron-select` and `paper-menu-button`. this is because, by convention, we're meant to give it a child of `paper-menu`, which _does_ implement iron-selectable (we are relying on bubbling).

also, should the analyser be detecting `fire(...)` calls to populate the events array?

43081j:
@rictic @justinfagnani maybe in the analyser, we could detect undocumented events by checking `this.fire(...)` calls. We could warn in the analyser (or linter) then for them being undocumented, and use them in places like this PR too.

I suppose we could just give `PolymerElementScanner` an `enterExpressionStatement`? Check that the callee's type is `ThisExpression` and name is `fire` by the looks of it, then somehow build a (rather empty) `ScannedEvent` from it using the node's first argument value. Though not sure what the ScannedEvent's sourceRange/node should be if we did that...

FredKSchott:
Closing this PR as abandoned. This happens if there hasn't been any activity on a PR in a long time.

@43081j if you are still interested in getting this PR merged: please rebase & respond to comments, and then ping me to reopen and I'll be happy to take a look.

PR #48, opened 2017-03-07
Calls in databindings must be functions

### LINKED ISSUE #29: Observers must be functions
Port from polylint. See observerNotFunction here: https://github.com/PolymerLabs/polylint/blob/master/src/linters.ts#L505

rictic:

Fixes #29 

---
This change is [Reviewable](https://reviewable.io/reviews/polymer/polymer-linter/48)

### DIFF
--- a/src/polymer/databinding-calls-must-be-functions.ts
+++ b/src/polymer/databinding-calls-must-be-functions.ts
@@ -107,7 +107,7 @@ export class DatabindingCallsMustBeFunctions extends Rule {
       }
       const potentialMethodNames = new Set(
           element.properties
-              .filter((p) => !definitelyNotMethodTypes.has(p.type || ''))
+              .filter((p) => !p.type || !definitelyNotMethodTypes.has(p.type))
               .map((p) => p.name)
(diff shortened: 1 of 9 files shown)


src/polymer/databinding-calls-must-be-functions.ts
```diff
@@ -0,0 +1,131 @@
...
+        }
+      }
+
+      // Early exit if there's nothing to check.
+      if (mustBeMethods.length === 0) {
+        continue;
+      }
+      const potentialMethodNames = new Set(
+          element.properties
+              .filter((p) => !definitelyNotMethodTypes.has(p.type || ''))
```
usergenic: maybe `(p) => !p.type || !definitelyNotMethodTypes.has(p.type)` instead here.

usergenic (approved):
LGTM (with optional nit for abuse of my logic receptors "filter where not definitelyNotMethods has nothing")

src/polymer/databinding-calls-must-be-functions.ts
rictic: I don't not undisagree.

Done.

PR #53, opened 2017-03-13
custom elements must call super.callback()

### LINKED ISSUE #52: Warn when super must be called but it isn't
Polymer 2 elements that override methods like `connectedCallback`, you must call `super.connectedCallback()` or you get weird, difficult to debug errors.

This came up in `#2.0-preview` on slack.

rictic:
Fixes #52 

---
This change is [Reviewable](https://reviewable.io/reviews/polymer/polymer-linter/53)

### DIFF
--- a/src/polymer/call-super-in-callbacks.ts
+++ b/src/polymer/call-super-in-callbacks.ts
@@ -85,28 +85,32 @@ export class CallSuperInCallbacks extends Rule {
             const sourceRange =
                 parsedDocumentContaining.sourceRangeForNode(method.key)!;
             if (method.kind === 'constructor') {
-              warnings.push(
-                  {
-                    code: this.code,
-                    severity: Severity.ERROR, sourceRange,
-                    message: stripWhitespace(`
-                  ${elementLike.className || elementLike.tagName} must call
-                  super() in its constructor because ES6 requires it.
-                `)
-                  });
-            } else {
-              const overrides = elementLike instanceof ElementMixin ?
-                  'may override a' :
-                  'overrides the';
               warnings.push({
-                code: this.code,
-                severity: Severity.WARNING, sourceRange,
+                code: 'call-super-in-constructor',
+                severity: Severity.ERROR, sourceRange,
                 message: stripWhitespace(`
-                  ${getName(elementLike)} may need to call
-                  super.${methodName}() because it ${overrides}
-                  version in ${classThatRequiresSuper}.
+                  ES6 requires super() in constructors with superclasses.
                 `)
               });
+            } else {
+              let message;
+              let code;
+              if (elementLike instanceof ElementMixin) {
+                code = 'call-super-in-mixin-callbacks';
+                message = stripWhitespace(`
+                    This method should conditionally call super.${methodName}()
+                    because a class ${getName(elementLike, 'this mixin')} is
+                    applied to may also define ${methodName}.`);
+              } else {
+                code = this.code;
+                message = stripWhitespace(`
+                    You may need to call super.${methodName}() because
+                    ${getName(elementLike, 'this class')} extends
+                    ${classThatRequiresSuper}, which defines ${methodName} too.
+                `);
+              }
+              warnings.push(
+                  {severity: Severity.WARNING, code, sourceRange, message});
             }
           }
         }
@@ -251,11 +255,12 @@ function getMethodDefiner(
   }
 }
 
-function getName(elementLike: Element|ElementMixin) {
+function getName(elementLike: Element|ElementMixin, fallback?: string) {
   if (elementLike instanceof Element) {
-    return elementLike.className || elementLike.tagName || 'Unknown Element';
+    return elementLike.className || elementLike.tagName || fallback ||
+        'Unknown Element';
   } else {
-    return elementLike.name || 'Unknown Mixin';
+    return elementLike.name || fallback || 'Unknown Mixin';
   }
 }
 
--- a/src/polymer/call-super-in-callbacks.ts
+++ b/src/polymer/call-super-in-callbacks.ts
@@ -14,7 +14,7 @@
 
 import * as estraverse from 'estraverse';
 import * as estree from 'estree';
-import {Document, isPositionInsideRange, ParsedJavaScriptDocument, Severity, SourceRange, Warning} from 'polymer-analyzer';
+import {Document, isPositionInsideRange, ParsedDocument, Severity, SourceRange, Warning} from 'polymer-analyzer';
 
 import {registry} from '../registry';
 import {Rule} from '../rule';
@@ -113,7 +113,7 @@ export class CallSuperInCallbacks extends Rule {
 
 function getParsedDocumentContaining(
     sourceRange: SourceRange|undefined,
-    document: Document): ParsedJavaScriptDocument|undefined {
+    document: Document): ParsedDocument<any, any>|undefined {
   if (!sourceRange) {
     return undefined;
   }
@@ -128,8 +128,7 @@ function getParsedDocumentContaining(
     }
   }
   mostSpecificDocument = mostSpecificDocument || document;
-  // TODO(rictic): export ParsedDocument from analyzer.
-  return mostSpecificDocument.parsedDocument as any;
+  return mostSpecificDocument.parsedDocument;
 }
 
 function getClassBody(astNode?: estree.Node|null): undefined|estree.ClassBody {
--- a/src/polymer/call-super-in-callbacks.ts
+++ b/src/polymer/call-super-in-callbacks.ts
@@ -168,7 +168,7 @@ function mustCallSuper(
   }
   // ElementMixins should always conditionally call super in callbacks.
   if (elementLike instanceof ElementMixin) {
-    return `the mixin's given superclass`;
+    return `some of the classes this mixin may be applied to`;
   }
   // Did the element's super class define the method?
   if (elementLike.superClass) {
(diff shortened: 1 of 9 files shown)


justinfagnani:
@rictic I think this might need a little more specification.

Elements with a known super class probably want to call `super.X()` if `X` is know to be defined at compile time, whether or not the elements are vanilla.

Mixins however should probably always do a guarded call, like

```
if (super.X) super.X();
```

And of course there are cases where you want to not can super on purpose.

rictic:
I agree with all of those objections, but as I understand them they only apply to a generalization of this lint pass.

Every Polymer 2.0 element should call super in `connectedCallback`, `disconnectedCallback`, `attributesChangedCallback`, and the constructor or they'll get undefined behavior that may change from minor version to minor version of Polymer. I think that's always true.

The generalized form is much harder to define and write. I think it'd involve something like an `@mustCallSuperInOverrides` (🚲🏡)  jsdoc annotation on a method.

I think this specific form is valuable, as people have been running into this error in the wild, and the failure mode is a mysterious and hard to track down. The element just doesn't stamp out its template.

justinfagnani:
I think we need to be careful with the phrase "every Polymer element", because in 2.0 that doesn't mean that you extend from `Polymer.Element` or even have `ElementMixin` applied. Mainly what being a Polymer element means right now is that we read `is` and `properties`. So a Polymer element might not have a callback defined. And a vanilla element may extend a base class or apply a mixin with a callback defined.

We can actually check whether we see a callback and lint based on that with no fork on polymer vs vanilla.

rictic:
Oh, you mean like looking up the inheritance graph to see if anything up there defines any of the callbacks and assuming that they're required if defined? Yeah that makes sense to me. I'll take a shot.

rictic:
@justinfagnani Done, PTAL.

I'd recommend starting from the tests, then going to the implementation. (same with all these lint passes really)

src/polymer/call-super-in-callbacks.ts
```diff
@@ -0,0 +1,221 @@
...
+  `).trim();
+
+  constructor() {
+    super();
+  }
+
+  async check(document: Document) {
+    const warnings: Warning[] = [];
+
+    for (const element of document.getByKind('polymer-element')) {
```
justinfagnani: Why is this polymer-element, and not element? And should this also contain element-mixin?

src/polymer/call-super-in-callbacks.ts
```diff
@@ -0,0 +1,221 @@
...
+      // TODO(rictic): methods should have astNodes, that would make this
```
justinfagnani: Add an issue on the analyzer

src/polymer/call-super-in-callbacks.ts
```diff
@@ -0,0 +1,221 @@
...
+            const nameOfMethod = method.kind === 'constructor' ?
+                'constructor' :
+                `${methodName} method override`;
+            warnings.push({
+              code: this.code,
+              severity: Severity.ERROR,
+              sourceRange:
+                  parsedDocumentContaining.sourceRangeForNode(method.key)!,
+              message: stripWhitespace(`
+                Elements that extend Polymer.Element must call
```
justinfagnani: Remove Polymer.Element add class or mixin that declares method

src/polymer/call-super-in-callbacks.ts
```diff
@@ -0,0 +1,221 @@
...
+      if (node.type === 'ClassDeclaration' || node.type === 'ClassExpression') {
+        classBody = node.body;
+        return estraverse.VisitorOption.Break;
+      }
+    }
+  });
+  return classBody;
+}
+
+function mustCallSuper(
```
justinfagnani: You might want to return the feature that defined the method

src/polymer/call-super-in-callbacks.ts
```diff
@@ -0,0 +1,221 @@
...
+      return true;
+    }
+  }
+
+  return anyMixinDefinesMethod(element, methodName, document, true);
+}
+
+function doesCallSuper(method: estree.MethodDefinition, methodName: string) {
+  const statements = method.value.body.body;
+  const superCallTargets = statements.map((s) => {
```
justinfagnani: It looks like this only finds top level calls. As a test for a mixin that conditionally calls the super method.

justinfagnani (changes requested):

src/polymer/call-super-in-callbacks.ts
rictic: Now querying for elements and element mixins.

Now blocked on a tiny PR in the analyzer: https://github.com/Polymer/polymer-analyzer/pull/561

Tests will fail until that's released.

src/polymer/call-super-in-callbacks.ts
rictic: Done

src/polymer/call-super-in-callbacks.ts
rictic: Rewrote, PTAL

src/polymer/call-super-in-callbacks.ts
rictic: +1.

Returned a string representation.

src/polymer/call-super-in-callbacks.ts
rictic: Yeah, I was overly narrow. Changed to being overly broad. Added test.

src/polymer/call-super-in-callbacks.ts
```diff
@@ -0,0 +1,268 @@
...
+    elementLike: Element, methodName: string, document: Document): (string|
+                                                                    undefined) {
+  // TODO(rictic): look up the inheritance graph for a jsdoc tag that describes
+  //     the method as needing to be called?
+  if (!methodsThatMustCallSuper.has(methodName)) {
+    return;
+  }
+  // ElementMixins should always conditionally call super in callbacks.
+  if (elementLike instanceof ElementMixin) {
+    return `the mixin's given superclass`;
```
justinfagnani: I don't quite read this string. A mixin doesn't have a superclass.

src/polymer/call-super-in-callbacks.ts
```diff
@@ -0,0 +1,268 @@
...
+    if (source) {
+      return source;
+    }
+  }
+  for (const mixinReference of elementLike.mixins) {
+    // TODO(rictic): once we have a representation of a Class this should be
+    //   something like `document.getById('class')` instead.
+    const mixin = onlyOrNone(
+        document.getById('element-mixin', mixinReference.identifier));
+    // TODO(rictic): if mixins had their own mixins pre-mixed in we wouldn't
```
justinfagnani: file issue?

src/polymer/call-super-in-callbacks.ts
```diff
@@ -0,0 +1,268 @@
...
+    document: Document,
+    skipLocalCheck: boolean): string|undefined {
+  if (!skipLocalCheck) {
+    const source = definesMethod(elementLike, methodName);
+    if (source) {
+      return source;
+    }
+  }
+  for (const mixinReference of elementLike.mixins) {
+    // TODO(rictic): once we have a representation of a Class this should be
```
justinfagnani, line 225: file issue?

src/polymer/call-super-in-callbacks.ts
```diff
@@ -0,0 +1,268 @@
...
+    const cause =
+        mixin && anyMixinDefinesMethod(mixin, methodName, document, false);
+    if (cause) {
+      return cause;
+    }
+  }
+  return;
+}
+
+function definesMethod(
```
justinfagnani: doc that because we copy down properties this finds methods up the inheritance chain

src/polymer/call-super-in-callbacks.ts
rictic: I'm trying to come up with something better than `mixedin` :)

How about `a class this mixin extends`?

justinfagnani: But there is no such class. What entity are you referring to here?

rictic: e.g. `superClass` in this example:

```javascript
function myMixin(superClass) {
  return class extends superClass { /* ... */ };
}
```

justinfagnani: That's just a parameter, so it can't really be the source of a method no?

rictic: Given the assumptions:

  1. if your superclass defines a custom element callback and you override it, then you must call super.thatCallback() in that override
  2. a mixin may be mixed into any class (or any element class, same diff)

then it follows that any mixin that implements a custom element callback must conditionally call super.thatCallback()

src/polymer/call-super-in-callbacks.ts
rictic: Done

rictic (commented):
Comments addressed except for the recent long mixin thread which got messed up because github.

rictic:
Sidenote but we should really try reviewable. I know it's got an overcomplicated interface, but it seems like it is doing a better job of representing what's going on in a code review of a PR.

rictic:
Ok, new version of analyzer is used now, tests pass, PTAL.

src/polymer/call-super-in-callbacks.ts
justinfagnani: Ok, if the resultant message reads something like we talked about it's fine.

justinfagnani (approved):

PR #55, opened 2017-03-15
prepare for 1.0.0 release

rictic:

---
This change is [Reviewable](https://reviewable.io/reviews/polymer/polymer-linter/55)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,65 @@
 # Polymer Linter
 
-Catch errors in your Polymer project before even running your code.
+Analyze your Polymer project for errors.
 
-## Work in progress
+## Usage
 
-Polymer Linter is being developed to replace [polylint][1], but is not yet complete.  For a working linter, use [polylint][1].
+The linter can be run in one of two ways, from the command line, or through our editor plugins.
 
-[1]: https://www.github.com/polymerlabs/polylint
+First though, it needs to be configured. Create a `polymer.json` file at the root of your project and ensure it has a "lint" field like so:
+
+```json
+{
+  "lint": {
+    "rules": ["polymer-2-hybrid"]
+  }
+}
+```
+
+You can specify either individual rules, or rule collections. See `polymer lint --help` for a full listing, or see [collections.ts](src/collections.ts) for details.
+
+### From the command line
+
+polymer-lint is run on the command line through Polymer CLI:
+
+```
+  npm install -g polymer-cli@next
+  cd my-project
+  polymer lint
+```
+
+```
+    <paper-button opan="{{nonono}}"></paper-button>
+                  ~~~~
+
+my-elem.html(9,24) warning [set-unknown-attribute] - paper-button elements do not have a property named opan. Consider instead:  open
+```
+
+### Inside your editor
+
+polymer-lint is also integrated into a number of editor plugins for instant as-you-type linting. See [https://github.com/Polymer/polymer-editor-service](the polymer editor service) for details.
+
+<!-- TODO(rictic): animated gif of editing text with linting here. -->
+
+## Writing your own lint rule
+
+The linter is built on top of the [polymer analyzer](https://github.com/Polymer/polymer-analyzer). A lint rule is given a `Document` object with an AST that can be walked, as well as the ability to query high level features of the document like imports and custom elements. From this is just has to return an array of warnings to display.
+
+For a simple example, see [behaviors-spelling](src/polymer/behaviors-spelling.ts), which implements checks for the commonwealth spelling of the property `behaviors` on a Polymer element.
+
+You'll then need to import your rule from [rules.ts](src/rules.ts), and you'll probably want to add your rule to one or more [rule collections](src/collections.ts).
+
+### Testing
+
+Testing is straightforward and robust. Test code is in `src/test` and fixtures live in `test/`. We have facilities for ensuring that your warnings pinpoint the issue to the exact right section of code, as well as ensuring that your rule was registered correctly.
+
+See [behaviors-spelling_test.ts](src/test/polymer/behaviors-spelling_test.ts) for an example.
+
+### Help writing lint rules
+
+More lint rules are very much welcome! We're happy to answer questions. We've got a very welcoming community, come join us in the [#tools channel on slack](https://polymer.slack.com/messages/tools)! ([invites emailed automatically here](https://polymer-slack.herokuapp.com/))
+
+## Previous Version
+
+Polymer Linter is the successor to [polylint](https://www.github.com/polymerlabs/polylint).
+
--- a/README.md
+++ b/README.md
@@ -1,10 +1,10 @@
 # Polymer Linter
 
-Analyze your Polymer project for errors.
+A library for finding and reporting on suspicious patterns and errors in web projects.
 
 ## Usage
 
-The linter can be run in one of two ways: from the command line, or through our editor plugins.
+The Polymer Linter can be run from command line with the [Polymer CLI](https://github.com/Polymer/polymer-cli), or though IDE plugins, such as the [VS Code Plugin](https://github.com/Polymer/vscode-plugin).
 
 First though, it needs to be configured. Create a `polymer.json` file at the root of your project and ensure it has a "lint" field like so:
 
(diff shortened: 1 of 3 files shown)


rictic:
The animated gif doesn't display right in the preview section, but you can see it here: https://github.com/Polymer/polymer-linter/blob/00443cbfdbbfb287832afe3e2426900d2344a338/README.md

web-padawan:
@rictic Are you going to add a CLI to allow using linter without installing `polymer-cli`?

rictic:
Current plan is no. polymer-cli does a lot of the work needed for a CLI. It parses the project's polymer.json, does a good job of handling command line args, and has true integration tests for the CLI, all of which would be somewhat unfortunate to duplicate over here.

Why do you ask? Is there something about `polymer-cli` that doesn't work for your workflow?

web-padawan:
Installing `polymer-cli` means lots of dependencies, and some of them might not be necessary, e. g. if we don't use `init`, why do we have to install `yeoman` stuff? 

Another thing is that projects using `polymer-build` and `wct` separately may have pinned versions (e. g. we are still using `wct` 4.3.1 due to `istanbul` bug, which was not resolved until 6.0.0-prerelease). 

So it would be nice to also have an opportunity to install `polymer-linter` separately.

rictic:
That's fair. We know that we need to get `polymer-cli`'s size down. For the wct versioning issue, if you install a different wct version locally in a project, `polymer-cli` will defer to it, so you can have the latest polymer-cli installed globally, and a targeted version of wct installed on a per-project basis.

We'll be listening to feedback, and if there's enough demand for a standalone linter command I think we could provide one.

usergenic (approved):
Enthusiastic LGTM

README.md
```diff
@@ -1,9 +1,67 @@
 # Polymer Linter
 
-Catch errors in your Polymer project before even running your code.
+Analyze your Polymer project for errors.
 
-## Work in progress
+## Usage
 
-Polymer Linter is being developed to replace [polylint][1], but is not yet complete.  For a working linter, use [polylint][1].
+The linter can be run in one of two ways: from the command line, or through our editor plugins.
```
FredKSchott: Per the PR discussion, you may want to specify here that it can be run from within `polymer-cli`, and does not come with it's own CLI interface.

Can't you also run the linter via the JS interface? For someone who's reading the repo README, that might be worth mentioning without removing the nod to editor plugins.

Here's how i'd rewrite, feel free to use or modify:

```
The linter can be run on the command line (with the [Polymer CLI]) or JavaScript. We also support [several editor plugins] for live-linting in your text editor.
```

README.md
```diff
@@ -1,9 +1,67 @@
...
+
+![Animated gif of the VSCode with the linter running.](https://cloud.githubusercontent.com/assets/1659/23933285/ad63eb62-08fa-11e7-819b-641bf83cf9c6.gif)
+
+## Writing your own lint rule
+
+The linter is built on top of the [polymer analyzer](https://github.com/Polymer/polymer-analyzer). A lint rule is given a `Document` object with an AST that can be walked, as well as the ability to query high level features of the document like imports and custom elements. From this is just has to return an array of warnings to display.
+
+For a simple example, see [behaviors-spelling](src/polymer/behaviors-spelling.ts), which implements a check for the commonwealth spelling of the property `behaviors` on a Polymer element.
+
+You'll then need to import your rule from [rules.ts](src/rules.ts), and you'll probably want to add your rule to one or more [rule collections](src/collections.ts).
```
FredKSchott: Hmm, you kind of lost me here. Does this mean there is no way to add rules to the linter without checking out and modifying the repo directly? If so, you should make that clear at the beginning of the section.

README.md
```diff
@@ -1,9 +1,67 @@
...
+
+### Testing
+
+Testing is straightforward and robust. Test code is in `src/test` and fixtures live in `test/`. We have facilities for ensuring that your warnings pinpoint the issue to the exact right section of code, as well as ensuring that your rule was registered correctly.
```
FredKSchott: This section is nicely written but ultimately doesn't tell me much about testing. I'd cut it down to just the basic info: 
- run the tests with `npm run test`
- tests exist in src/test/
- fixtures exist in test/
- see `behaviors-spelling_test.ts` for a good example of how to write your own test

README.md
```diff
@@ -1,9 +1,67 @@
...
+
+See [behaviors-spelling_test.ts](src/test/polymer/behaviors-spelling_test.ts) for an example.
+
+### Help writing lint rules
```
FredKSchott: If this is a call to action, no harm in giving it a little personality. Something like: "Help Wanted: More Lint Rules" or even "New Lint Rules Welcome!"

README.md
```diff
@@ -1,9 +1,67 @@
...
+
+More lint rules are very much welcome! We're happy to answer questions. We've got a very welcoming community, come join us in the [#tools channel on slack](https://polymer.slack.com/messages/tools)! ([invites emailed automatically here](https://polymer-slack.herokuapp.com/))
+
+## Previous Version
```
FredKSchott: Same here re: personality. "Looking for Polylint?"

FredKSchott (approved):
Note: All comments are purely personal opinion and non-binding :) Overall the content is great

README.md
rictic: I'm comfortable with how this part is written. We mention how to run it from the command line a couple lines down with all the details

README.md
rictic: Clarified, here, as it's a temporary limitation.

README.md
rictic: +1

README.md
rictic: +1

README.md
FredKSchott: You may be overestimating how much people read vs. skim. As written, I assume that this package gives me something that can be run directly on the command line.

README.md
```diff
@@ -1,9 +1,69 @@
 # Polymer Linter
 
-Catch errors in your Polymer project before even running your code.
+Analyze your Polymer project for errors.
```
justinfagnani: Feels a little vague. Looking around at other descriptions, how about something like:

"A library for finding and reporting on suspicious patterns and errors in web projects."

README.md
justinfagnani: I would change this to "The Polymer Linter can be run from command line with the [Polymer CLI], or though IDE plugins, such as the [VS Code Plugin]."

README.md
```diff
@@ -1,9 +1,69 @@
...
+
+```json
+{
+  "lint": {
+    "rules": ["polymer-2-hybrid"]
+  }
+}
+```
+
+You can specify either individual rules, or rule collections. See `polymer lint --help` for a full listing, or see [collections.ts](src/collections.ts) for details.
```
justinfagnani: I don't think we should be sending end users into code. We should should separate the docs between CLI/IDE users and library users. Can be TODO for now

README.md
```diff
@@ -1,9 +1,69 @@
...
+my-elem.html(9,24) warning [set-unknown-attribute] - paper-button elements do not have a property named opan. Consider instead:  open
+```
+
+### Inside your editor
+
+polymer-lint is also integrated into a number of editor plugins for instant as-you-type linting. See [the polymer editor service](https://github.com/Polymer/polymer-editor-service) for details.
+
+![Animated gif of the VSCode with the linter running.](https://cloud.githubusercontent.com/assets/1659/23933285/ad63eb62-08fa-11e7-819b-641bf83cf9c6.gif)
+
+## Writing your own lint rule
```
justinfagnani: Let's change the top heading of this section to something like "Extending and Contributing"

README.md
rictic: Ok, I'm outvoted :) done

README.md
rictic: Done

README.md
rictic: Ideally if they've got an IDE plugin we could assist them there in choosing rules.

Removed the reference to the source here.

README.md
rictic: Done

FredKSchott (approved):
LGTM

rictic:
I like that M. A lot of pep in that one's step.

justinfagnani (approved):

PR #59, opened 2017-03-17
Standardize tests

rictic:
- refer to rules only using the registry
- add tests for trivial and noop cases
- don't export rule classes, use the registry

 - [x] CHANGELOG.md not updated, internal change

---
This change is [Reviewable](https://reviewable.io/reviews/polymer/polymer-linter/59)


usergenic (approved):
LGTM

PR #61, opened 2017-03-20
When linting a package, don't warn about dependencies

rictic:

---
This change is [Reviewable](https://reviewable.io/reviews/polymer/polymer-linter/61)

### DIFF
--- a/src/html/undefined-elements.ts
+++ b/src/html/undefined-elements.ts
@@ -41,15 +41,21 @@ export class UndefinedElements extends HtmlRule {
         // HACK. Filed as https://github.com/Polymer/polymer-analyzer/issues/507
         continue;
       }
+      if (!ref.astNode) {
+        continue;
+      }
       const el = document.getById(
           'element', ref.tagName, {imported: true, externalPackages: true});
 
       if (el.size === 0) {
+        const sourceRange = parsedDocument.sourceRangeForStartTag(ref.astNode);
+        if (!sourceRange) {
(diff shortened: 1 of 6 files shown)


src/html/undefined-elements.ts
```diff
@@ -41,15 +41,21 @@ export class UndefinedElements extends HtmlRule {
...
       }
+      if (!ref.astNode) {
+        continue;
+      }
       const el = document.getById(
           'element', ref.tagName, {imported: true, externalPackages: true});
 
       if (el.size === 0) {
+        const sourceRange = parsedDocument.sourceRangeForStartTag(ref.astNode);
+        if (!sourceRange) {
```
rictic, line 55: It's a bit sloppy to have these unrelated changes here, but I was seeing some weird errors from these occasionally being undefined.

src/html/undefined-elements.ts
```diff
@@ -41,15 +41,21 @@ export class UndefinedElements extends HtmlRule {
         // HACK. Filed as https://github.com/Polymer/polymer-analyzer/issues/507
         continue;
       }
+      if (!ref.astNode) {
```
FredKSchott: is this a hack like above? Please comment if so

FredKSchott (approved):

src/html/undefined-elements.ts
rictic: :+1: Added comment

PR #64, opened 2017-03-24
[Prototype]: `lint --fix` design 

rictic:
I'd recommend starting by reading the types and their documentation in src/warning.ts, then the tests, then the rest of the implementation.

---
This change is [Reviewable](https://reviewable.io/reviews/polymer/polymer-linter/64)


FredKSchott (commented):
Overall +1, I really like the ability to attach the necessary edit to each warning. That looks really slick and won't require a whole new way of thinking about linting vs. fixing. Seriously, nice work on this @rictic.

Have you thought about deferring the actual fix logic (the call to`insertIntoMethod()`) until applyEdits is called? It's hard to eyeball the performance impact, but It looks like a significant amount of work to perform for each warning on a lint run. 

Also, what is the user flow if there are incompatible edits? Do we tell the user to run lint again? Do we automatically re-run `lint --fix` until all errors are fixed?

rictic:
Yeah, I'd like to defer the work. We could keep the API by making it a memoized getter.

I worried about this for a while with generating sourcemaps for warnings too, some of that work is significant. I've got a PR I'll send over that converts Warning from an interface to a class, which will make it easier for us to do this sort of lazy generation I think. I suppose we could define a getter that does lazy computation on an object literal though…

In general my thinking has been that in typical use we won't be generating tons of warnings, and that in situations where we are, it's ok to spend some time per-warning.

TimvdLippe:
@rictic As #105 has been merged, I think this one can be closed?

rictic:
+1, closing

PR #81, opened 2017-05-08
Update analyzer to 2.0.0 alpha.41

usergenic:
- Essentially a straightforward update to latest analyzer with a couple type changes in the `call-super-in-callbacks` `mustCallSuper()` signature. 

---
This change is [Reviewable](https://reviewable.io/reviews/polymer/polymer-linter/81)

### DIFF
--- a/src/polymer/call-super-in-callbacks.ts
+++ b/src/polymer/call-super-in-callbacks.ts
@@ -39,9 +39,10 @@ class CallSuperInCallbacks extends Rule {
   async check(document: Document) {
     const warnings: Warning[] = [];
 
-    const elementLikes =
-        Array.from(document.getFeatures({kind: 'element'}))
-            .concat(Array.from(document.getFeatures({kind: 'element-mixin'})));
+    const elementLikes = new Array<Element|ElementMixin>(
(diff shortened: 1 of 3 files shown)


src/polymer/call-super-in-callbacks.ts
```diff
@@ -39,9 +39,10 @@ class CallSuperInCallbacks extends Rule {
   async check(document: Document) {
     const warnings: Warning[] = [];
 
-    const elementLikes =
-        Array.from(document.getFeatures({kind: 'element'}))
-            .concat(Array.from(document.getFeatures({kind: 'element-mixin'})));
+    const elementLikes = new Array<Element|ElementMixin>(
+        ...document.getFeatures({kind: 'element'}),
```
rictic, line 43: consider changing this to just `'class'`, here and elsewhere

ok to punt this as well though.

rictic (approved):
One optional improvement. LGTM so long as tests pass.

src/polymer/call-super-in-callbacks.ts
usergenic: Punting because I don't understand what changes to `'class'` and why...

PR #109, opened 2017-10-14
Support converting declarations of <content> to <slot> in addition to usages.

rictic:

---
This change is [Reviewable](https://reviewable.io/reviews/polymer/polymer-linter/109)

### DIFF
--- a/src/polymer/content-to-slot-declarations.ts
+++ b/src/polymer/content-to-slot-declarations.ts
@@ -1,0 +1,113 @@
+/**
+ * @license
+ * Copyright (c) 2017 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at
+ * http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import * as dom5 from 'dom5';
+import * as parse5 from 'parse5';
+import {treeAdapters} from 'parse5';
+import {Document, ParsedHtmlDocument, Severity} from 'polymer-analyzer';
+
+import {HtmlRule} from '../html/rule';
+import {registry} from '../registry';
+import {FixableWarning} from '../warning';
+
+import stripIndent = require('strip-indent');
+
+
+const p = dom5.predicates;
+
+class ContentToSlotDeclarations extends HtmlRule {
+  code = 'content-to-slot-declarations';
+  description = stripIndent(`
+      Warns when using <content> instead of Shadow Dom v1's <slot> element.
+  `).trim();
+
+  async checkDocument(parsedDocument: ParsedHtmlDocument, document: Document) {
+    const warnings: FixableWarning[] = [];
+
+    this.convertDeclarations(parsedDocument, document, warnings);
+
+    return warnings;
+  }
+
+  convertDeclarations(
+      parsedDocument: ParsedHtmlDocument, document: Document,
+      warnings: FixableWarning[]) {
+    for (const element of document.getFeatures({kind: 'polymer-element'})) {
+      const domModule = element.domModule;
+      if (!domModule) {
+        continue;
+      }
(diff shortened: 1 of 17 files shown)


rictic:
Hm, on reflection, we probably shouldn't be doing this automatically... probably.

Need to think about this a bit more I think.

rictic:
Updated to not apply this automatically. Should follow up and give a better way to warn without automatically fixing.

src/polymer/content-to-slot-declarations.ts
```diff
@@ -0,0 +1,117 @@
...
+        const warning = new FixableWarning({
+          code: 'content-to-slot-declaration',
+          message:
+              `<content> tags are part of the deprecated Shadow Dom v0 API. ` +
+              `Replace with a <slot> tag.`,
+          parsedDocument,
+          severity: Severity.WARNING,
+          sourceRange: parsedDocument.sourceRangeForStartTag(contentNode)!
+        });
+        const attrs = [...contentNode.attrs];
```
valdrinkoshi: got lost on what's happening in this block (here to line 85). Comments/split it in a separate function maybe?

test/content-to-slot-declarations/before-fixes.html
```diff
@@ -0,0 +1,13 @@
+<dom-module id="foo-bar">
+  <template>
+    <content></content>
```
valdrinkoshi: nit: move as last `<content>`

test/content-to-slot-declarations/after-fixes.html
```diff
@@ -0,0 +1,13 @@
+<dom-module id="foo-bar">
+  <template>
+    <slot></slot>
```
valdrinkoshi: nit: move as last `<slot>`

src/polymer/content-to-slot-declarations.ts
```diff
@@ -0,0 +1,117 @@
...
+        const selectorAttr = attrs.find((a) => a.name === 'select');
+        const selector = selectorAttr && selectorAttr.value;
+        let slotName = null;
+        if (selector) {
+          slotName = slotNameForSelector(selector);
+          while (slots.has(slotName)) {
+            slotName += '-unique-suffix';
```
valdrinkoshi: could you add a test for this? e.g.
Before:
```html
<content select="#foo">
  Foo
</content>
<content select=".foo">
  Foo
</content>
```
After:
```html
<slot name="foo" old-content-selector="#foo">
  Foo
</slot>
<slot name="foo-unique-suffix" old-content-selector=".foo">
  Foo
</slot>

File: src/polymer/content-to-slot-declarations.ts [hunk 5]
```diff
@@ -0,0 +1,117 @@
...
+          parsedDocument,
+          severity: Severity.WARNING,
+          sourceRange: parsedDocument.sourceRangeForStartTag(contentNode)!
+        });
+        const attrs = [...contentNode.attrs];
+        const selectorAttr = attrs.find((a) => a.name === 'select');
+        const selector = selectorAttr && selectorAttr.value;
+        let slotName = null;
+        if (selector) {
+          slotName = slotNameForSelector(selector);
```
valdrinkoshi: Selectors like ".baz .foo" should not be fixable, e.g. 
```html
<content select=".baz .foo">
```
currently gets updated to:
```html
<slot name="baz" old-content-selector=".baz .foo">
```

src/polymer/content-to-slot-declarations.ts
rictic: +1, done

src/polymer/content-to-slot-declarations.ts
rictic: Done

test/content-to-slot-declarations/after-fixes.html
rictic: Done

test/content-to-slot-declarations/before-fixes.html
rictic: Done

src/polymer/content-to-slot-declarations.ts
rictic: Hm, yeah good point! 

I agree, but I think that's something to add to the usages PR, as that sort of `<content>` element still needs to be upgraded to a `<slot>`, and I think it makes sense to preserve the old content selector as it is.

I'll add a check to the usages PR, I wanted to do a better job with selectors anyways.

valdrinkoshi (approved):
LGTM 👍

PR #112, opened 2017-10-27
Support content->slot conversions of arbitrary cooperating elements.

rictic:

---
This change is [Reviewable](https://reviewable.io/reviews/polymer/polymer-linter/112)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 - Support warning about and automatically fixing some simple uses of some 1.0 elements to their 2.0 usage styles.
   - To add this fixable lint warning for your own elements, simply add the `old-content-selector` attribute to your new `<slot>` element. Set its value to the value that your old `<content>` element had for its `select` attribute.
+    - e.g. if you had `<content select=".some-content">` and migrated to `<slot name="new-content" old-content-selector=".some-content">` then the linter will warn about and add `slot="new-content"` to children of your element that match `.some-content`.
+<slot old-content-selector=".some-content"></slot>>
 - Support warning about and automatically fixing element declarations that use `<content>` to use `<slot>` instead. This is a breaking change to an element,
 so it is not on by default.
   - It automatically adds the `old-content-selector` attribute to migrated `<slot>` elements, so uses can be automatically upgraded as well.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 - Support warning about and automatically fixing some simple uses of some 1.0 elements to their 2.0 usage styles.
+  - To add this fixable lint warning for your own elements, simply add the `old-content-selector` attribute to your new `<slot>` element. Set its value to the value that your old `<content>` element had for its `select` attribute.
 - Support warning about and automatically fixing element declarations that use `<content>` to use `<slot>` instead. This is a breaking change to an element,
 so it is not on by default.
+  - It automatically adds the `old-content-selector` attribute to migrated `<slot>` elements, so uses can be automatically upgraded as well.
 
 ## [2.1.0] - 2017-10-13
 - Warn for old-style @apply without parentheses, and var() with a fallback value of a bare css variable.
--- a/src/polymer/content-to-slot-usages.ts
+++ b/src/polymer/content-to-slot-usages.ts
@@ -25,12 +25,12 @@ import stripIndent = require('strip-indent');
 
 const p = dom5.predicates;
 
-const config =
+const staticConfig =
     new Map<string, Array<{predicate: dom5.Predicate, slot: string}>>();
 
 function addPredicate(
     tagname: string, slots: Array<{selector: string, slot: string}>) {
-  config.set(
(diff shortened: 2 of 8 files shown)


test/content-to-slot-usages/after-fixes.html
```diff
@@ -61,3 +61,27 @@
...
+  <div class="foo" slot="foo"></div>
+  <div class="bar" slot="bar"></div>
+  <div class="baz">
+    <div class="foo"></div>
+  </div>
+</foo-bar>
+
+<dom-module id="foo-bar">
+  <template>
+    <slot></slot>
```
valdrinkoshi: nit: this should be moved as last slot as it has no name (would distribute all content in here)

test/content-to-slot-usages/before-fixes.html
```diff
@@ -61,3 +61,27 @@
...
+  <div class="foo"></div>
+  <div class="bar"></div>
+  <div class="baz">
+    <div class="foo"></div>
+  </div>
+</foo-bar>
+
+<dom-module id="foo-bar">
+  <template>
+    <slot></slot>
```
valdrinkoshi: nit: this should be moved below as last slot as it has no name (would distribute all content in here)

src/polymer/content-to-slot-usages.ts
```diff
@@ -84,6 +84,35 @@ addPredicate(
     'paper-dropdown-menu',
     [{selector: '.dropdown-content', slot: 'dropdown-content'}]);
 
+function getContentDescriptors(tagName: string, document: Document):
```
valdrinkoshi: documentation needed 📄

src/polymer/content-to-slot-usages.ts
```diff
@@ -84,6 +84,35 @@ addPredicate(
...
+    Array<{slot: string, predicate: dom5.Predicate}>|undefined {
+  const descriptors = [];
+  const [element, ] = document.getFeatures({
+    kind: 'polymer-element',
+    id: tagName,
+    imported: true,
+    externalPackages: true
+  });
+  if (element && element.domModule) {
```
valdrinkoshi: what about elements in the main document? should that be a separate run?

rictic:
I added some more documentation, and added warnings for unsupported CSS expressions in `old-content-selector`, to handle the case pointed out here: https://github.com/Polymer/polymer-linter/pull/109#discussion_r147539419

The heavier processing of selectors made it seem worth it to cache results, so I added a weak map cache based on the Element object (which is immutable, so whenever it changes we'll get a different instance and thus a different entry in the weak map).

PTAL!

src/polymer/content-to-slot-usages.ts
```diff
@@ -172,27 +121,219 @@ class ContentToSlot extends HtmlRule {
...
+}
+
+class DescriptorError {
+  constructor(public readonly message: string, public readonly slot: Slot) {
   }
-  if (simpleSelector.startsWith('[') && simpleSelector.endsWith(']')) {
-    return p.hasAttr(simpleSelector.slice(1, -1));
+}
+
+type Result<Good, Bad> = {
```
valdrinkoshi, line 168: ..and the Ugly? 😁

CHANGELOG.md
```diff
@@ -7,8 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 - Support warning about and automatically fixing some simple uses of some 1.0 elements to their 2.0 usage styles.
+  - To add this fixable lint warning for your own elements, simply add the `old-content-selector` attribute to your new `<slot>` element. Set its value to the value that your old `<content>` element had for its `select` attribute.
```
valdrinkoshi, line 10: I'd add an example here, wdyt of something like
```html
<!-- <content select=".some-content"></content> -->
<slot old-content-selector=".some-content"></slot>

File: src/polymer/content-to-slot-usages.ts [hunk 7]
```diff
@@ -172,27 +121,219 @@ class ContentToSlot extends HtmlRule {
...
+function addPredicate(
+    tagname: string, slots: Array<{selector: string, slot: string}>) {
+  staticConfig.set(
+      tagname, slots.map((s) => ({
+                           predicate: elementSelectorToPredicate(s.selector),
+                           slot: s.slot
+                         })));
 }
 
+addPredicate(
```
valdrinkoshi, line 291: Probably material for a separate PR, but I'd define these predicates externally (e.g. a json file) and make the path configurable/extensible. This would allow element authors to share their upgrades to users of their elements.

valdrinkoshi (approved):

src/polymer/content-to-slot-usages.ts
rictic: I'd prefer to instead make changes to our elements to add the `old-content-selector` attribute and to remove this hard coded set. I added it in while I was initially hacking on this stuff. `old-content-selector` is better in every way.

I'll do that in a separate PR.

CHANGELOG.md
rictic: +1, wrote it out more explicitly

src/polymer/content-to-slot-usages.ts
rictic: :P

actually… when you've got a good result, a recoverable error, and an unrecoverable error… not a bad idea

PR #114, opened 2017-10-31
Automatically move styles into templates

rictic:

Moving them was fairly straightforward. Preserving their ordering and getting indentation correct often enough to be usable was the tricky part :)

Includes a change to the way that we apply edits. We now consider two replacements that insert text at the same place to be compatible.

Like most linter PRs, it's a bit easier to read the changes starting with the tests.

---
This change is [Reviewable](https://reviewable.io/reviews/polymer/polymer-linter/114)


src/html/move-style-into-template.ts
```diff
@@ -45,24 +61,164 @@ class MoveStyleIntoTemplate extends HtmlRule {
...
+  async checkDocument(parsedDocument: ParsedHtmlDocument, document: Document) {
+    const warnings: FixableWarning[] = [];
+    const domModules = document.getFeatures({kind: 'dom-module'});
+    for (const domModule of domModules) {
+      const template = (domModule.astNode.childNodes ||
+                        []).find((n) => n.tagName === 'template');
+      if (!template) {
+        continue;
+      }
+      const moduleChildren = domModule.astNode.childNodes || [];
```
valdrinkoshi: nit: can be moved above at line 68, and be used to find the template

src/html/move-style-into-template.ts
```diff
@@ -45,24 +61,164 @@ class MoveStyleIntoTemplate extends HtmlRule {
...
+            parsedDocument.sourceRangeForStartTag(template)!.end;
+
+        const styleIndentation = getIndentationInside(child);
+        const templateIndentation = getIndentationInside(
+            parse5.treeAdapters.default.getTemplateContent(template));
+        const clonedStyle = clone(child);
+        const contents = clonedStyle.childNodes;
+        if (styleIndentation === templateIndentation && contents != null &&
+            contents.every(dom5.isTextNode)) {
+          const additionalIndentation = '  ';
```
valdrinkoshi: This could be a useful common function, something like `indent(contents)` 💯

src/html/move-style-into-template.ts
```diff
@@ -45,24 +61,164 @@ class MoveStyleIntoTemplate extends HtmlRule {
...
+        const templateContentStart =
+            parsedDocument.sourceRangeForStartTag(template)!.end;
+
+        const styleIndentation = getIndentationInside(child);
+        const templateIndentation = getIndentationInside(
+            parse5.treeAdapters.default.getTemplateContent(template));
+        const clonedStyle = clone(child);
+        const contents = clonedStyle.childNodes;
+        if (styleIndentation === templateIndentation && contents != null &&
+            contents.every(dom5.isTextNode)) {
```
valdrinkoshi: why is this `contents.every(dom5.isTextNode)` check necessary?

src/html/move-style-into-template.ts
```diff
@@ -45,24 +61,164 @@ class MoveStyleIntoTemplate extends HtmlRule {
...
+          severity: Severity.WARNING, parsedDocument,
+          sourceRange: parsedDocument.sourceRangeForNode(linkInShadowDom)!
+        }));
+      }
     }
-    return warnings;
+
+    // Reverse here so that when we apply the fixes we move multiple
+    // styles into the template we preserve their order.
+    return warnings.reverse();
```
valdrinkoshi, line 165: 💯

src/html/move-style-into-template.ts
rictic: +1, done

src/html/move-style-into-template.ts
rictic: Maybe it's paranoid, but it's to be sure that there aren't any elements inside the `<style>` node.

I mean, there shouldn't ever be, but... just in case!

I moved the check into the new `addIndentation` function

src/html/move-style-into-template.ts
rictic: +1, extracted out

valdrinkoshi (approved):
👌

rictic:
Put getIndentationInside into utils and documented it.

PR #116, opened 2017-10-31
iron-flex-layout-import and iron-flex-layout-classes

valdrinkoshi:

Introducing 2 new rules for iron-flex-layout upgrades.
[iron-flex-layout v2 doesn't ship anymore](https://github.com/PolymerElements/iron-flex-layout/#changes-in-20) the long time deprecated `iron-flex-layout/classes/*.html` files.
These files were using the now deprecated /deep/ selector, and were styling the main document from an html import (feature not supported anymore in Chrome). 

The upgrade guide - see [Removing /deep/ and :shadow section](https://www.polymer-project.org/blog/2017-10-18-upcoming-changes.html) - suggests to:
- rename the import
- import the style modules in the element's shadow root

This PR implements 2 new rules to help achieve these 2 tasks.

### Limitations

1) Elements might be importing `iron-flex-layout/iron-flex-layout.html` and define themselves classes like: 
```html
<link rel="import" href="../iron-flex-layout/iron-flex-layout.html">
<dom-module id="my-element">
  <template>
    <style>
      .flex { @apply --layout-flex; }
    </style>
    <div class="flex">
```
In these cases `iron-flex-layout-classes` will still add the `style include="iron-flex"`, which wouldn't break styles, but wouldn't either be needed, resulting in:
```html
<link rel="import" href="../iron-flex-layout/iron-flex-layout.html">
<dom-module id="my-element">
  <template>
    <style include="iron-flex">
      .flex { @apply --layout-flex; }
    </style>
    <div class="flex">
```

2) There is no warning if classes are used but `iron-flex-layout/iron-flex-layout-classes.html` import is missing. As there is no prescribed way to import dependencies, it might be that the classes definitions are imported through a common import, e.g.
```html
<link rel="import" href="./commons.html">
<dom-module id="my-element">
  <template>
    <style include="iron-flex"></style>
    <div class="flex">
```
with `commons.html` being:
```html
<link rel="import" href="../iron-flex-layout/iron-flex-layout-classes.html">
```
This could be done in a separate PR, but requires all imports in the file to be parseable and we'd require to know the base path where to find iron-flex-layout - could be computed if there is already an import with iron-flex-layout/* href.

### Possible improvements
- When `iron-flex-layout-classes` is run on a non-polymer file which uses iron-flex-layout classes, it will erroneously warn and apply fix. Need a way to figure if the document is using polymer or not - maybe check the dependencies? 

---
This change is [Reviewable](https://reviewable.io/reviews/polymer/polymer-linter/116)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,10 +13,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Support warning about and automatically fixing element declarations that use `<content>` to use `<slot>` instead. This is a breaking change to an element,
 so it is not on by default.
   - It automatically adds the `old-content-selector` attribute to migrated `<slot>` elements, so uses can be automatically upgraded as well.
-- Support fixable warning about usage of deprecated `iron-flex-layout/classes/*.html` files via the new rule `iron-flex-layout-import`
-  - It automatically upgrades to import instead the `iron-flex-layout/iron-flex-layout-classes.html`
-- Support fixable warning about usage of iron-flex-layout classes without including the required style modules via the new rule `iron-flex-layout-classes`
-  - It automatically upgrades the element shadow root to include the iron-flex-layout style modules.
+- Support warning about and fixing usage of deprecated `iron-flex-layout/classes/*.html` files via the new rule `iron-flex-layout-import`
+  - It automatically upgrades to import the `iron-flex-layout/iron-flex-layout-classes.html` instead.
+- Support warning about and fixing usage of iron-flex-layout classes without including the required style modules via the new rule `iron-flex-layout-classes`
+  - It automatically upgrades the element template to include the iron-flex-layout style modules.
 
 ## [2.1.0] - 2017-10-13
 - Warn for old-style @apply without parentheses, and var() with a fallback value of a bare css variable.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,8 +13,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 so it is not on by default.
   - It automatically adds the `old-content-selector` attribute to migrated `<slot>` elements, so uses can be automatically upgraded as well.
 - Support automatic fixing of the warning where a `<style>` element is a direct child of a dom-module's `<template>`.
-- Support warning about and fixing usage of deprecated `iron-flex-layout/classes/*.html` files via the new rule `iron-flex-layout-import`
-  - It automatically upgrades to import the `iron-flex-layout/iron-flex-layout-classes.html` instead.
+- Support warning about and fixing usage of deprecated `iron-flex-layout/classes/*` files via the new rule `iron-flex-layout-import`
+  - It automatically adds, updates, or deletes the import `iron-flex-layout/iron-flex-layout-classes.html`.
 - Support warning about and fixing usage of iron-flex-layout classes without including the required style modules via the new rule `iron-flex-layout-classes`
   - It automatically upgrades the element template to include the iron-flex-layout style modules.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Support warning about and automatically fixing element declarations that use `<content>` to use `<slot>` instead. This is a breaking change to an element,
 so it is not on by default.
   - It automatically adds the `old-content-selector` attribute to migrated `<slot>` elements, so uses can be automatically upgraded as well.
+- Support fixable warning about usage of deprecated `iron-flex-layout/classes/*.html` files via the new rule `iron-flex-layout-import`
+  - It automatically upgrades to import instead the `iron-flex-layout/iron-flex-layout-classes.html`
+- Support fixable warning about usage of iron-flex-layout classes without including the required style modules via the new rule `iron-flex-layout-classes`
+  - It automatically upgrades the element shadow root to include the iron-flex-layout style modules.
 
 ## [2.1.0] - 2017-10-13
 - Warn for old-style @apply without parentheses, and var() with a fallback value of a bare css variable.
--- a/src/html/util.ts
+++ b/src/html/util.ts
@@ -15,6 +15,7 @@
 import * as dom5 from 'dom5';
 import cssWhat = require('css-what');
 import {ParsedHtmlDocument, SourceRange} from 'polymer-analyzer';
+import {Replacement} from '../warning';
 
 // Attributes that are on every HTMLElement.
 export const sharedAttributes = new Set([
@@ -262,4 +263,62 @@ export function removeTrailingWhitespace(
     }
   };
   return {range: newRange, replacementText: ''};
+}
+
+export function addAttribute(
+    parsedDocument: ParsedHtmlDocument,
+    node: dom5.Node,
+    attribute: string,
+    attributeValue: string): Replacement {
+  const tagRange = parsedDocument.sourceRangeForStartTag(node)!;
+  const range = {
+    file: tagRange.file,
+    start: {line: tagRange.end.line, column: tagRange.end.column - 1},
+    end: {line: tagRange.end.line, column: tagRange.end.column - 1}
+  };
+  const replacementText = ` ${attribute}="${attributeValue}"`;
+  return {replacementText, range};
+}
+
+export function prependContentInto(
+    parsedDocument: ParsedHtmlDocument,
+    node: dom5.Node,
+    replacementText: string): Replacement {
+  const tagRange = parsedDocument.sourceRangeForStartTag(node)!;
+  const range = {
+    file: tagRange.file,
+    start: {line: tagRange.end.line, column: tagRange.end.column},
+    end: {line: tagRange.end.line, column: tagRange.end.column}
+  };
+  return {replacementText, range};
+}
+
+export function insertContentAfter(
+    parsedDocument: ParsedHtmlDocument,
+    node: dom5.Node,
+    replacementText: string): Replacement {
+  const tagRange = parsedDocument.sourceRangeForNode(node)!;
+  const range = {
+    file: tagRange.file,
+    start: {line: tagRange.end.line, column: tagRange.end.column},
+    end: {line: tagRange.end.line, column: tagRange.end.column}
+  };
+  return {replacementText, range};
+}
+
+export function removeNode(
+    parsedDocument: ParsedHtmlDocument, node: dom5.Node): Replacement[] {
+  const parentChildren = node.parentNode!.childNodes!;
+  const prevNode = parentChildren[parentChildren.indexOf(node)! - 1];
+  const fix: Replacement[] = [];
+  if (prevNode && dom5.isTextNode(prevNode)) {
+    const trailingWhiteSpace =
+        removeTrailingWhitespace(prevNode, parsedDocument);
+    if (trailingWhiteSpace) {
+      fix.push(trailingWhiteSpace);
+    }
+  }
+  fix.push(
+      {replacementText: '', range: parsedDocument.sourceRangeForNode(node)!});
+  return fix;
 }
--- a/src/html/util.ts
+++ b/src/html/util.ts
@@ -14,6 +14,7 @@
 
 import * as dom5 from 'dom5';
 import cssWhat = require('css-what');
+import {ParsedHtmlDocument, SourceRange} from 'polymer-analyzer';
 
 // Attributes that are on every HTMLElement.
 export const sharedAttributes = new Set([
@@ -240,4 +241,25 @@ function attributeSelectorToPredicate(selector: cssWhat.Attribute):
   const never: never = selector.action;
   throw new Error(
       `Unexpected type of attribute matcher from CSS parser ${never}`);
+}
+
+export function removeTrailingWhitespace(
+    textNode: dom5.Node, parsedDocument: ParsedHtmlDocument) {
+  const prevText = dom5.getTextContent(textNode);
+  const match = prevText.match(/\n?[ \t]*$/);
+  if (!match) {
+    return;
+  }
+  const range = parsedDocument.sourceRangeForNode(textNode)!;
+  const lengthOfPreviousLine =
+      parsedDocument.newlineIndexes[range.end.line - 1] -
+      (parsedDocument.newlineIndexes[range.end.line - 2] || -1) - 1;
+  const newRange: SourceRange = {
+    ...range,
+    start: {
+      column: lengthOfPreviousLine,
+      line: range.end.line - 1,
+    }
+  };
+  return {range: newRange, replacementText: ''};
 }
--- a/src/html/util.ts
+++ b/src/html/util.ts
@@ -114,6 +114,7 @@ export const sharedProperties = new Set([
   'is',
 ]);
 
+/**
  * If the given node is a text node, add the given addition indentation to
  * each line of its contents.
  */
--- a/src/html/util.ts
+++ b/src/html/util.ts
@@ -12,6 +12,9 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
+import * as dom5 from 'dom5';
+import cssWhat = require('css-what');
+
 // Attributes that are on every HTMLElement.
 export const sharedAttributes = new Set([
   // From https://html.spec.whatwg.org/multipage/dom.html#htmlelement
@@ -110,3 +113,74 @@ export const sharedProperties = new Set([
 
   'is',
 ]);
+
+/**
+ * Converts a css selector into a dom5 predicate.
+ *
+ * This is intended for handling only selectors that match an individual element
+ * in isolation, it does throws if the selector talks about relationships
+ * between elements like `.foo .bar` or `.foo > .bar`.
+ */
+export function elementSelectorToPredicate(simpleSelector: string):
+    dom5.Predicate {
+  const parsed = cssWhat(simpleSelector);
+  // The output of cssWhat is two levels of arrays. The outer level are any
+  // selectors joined with a comma, so it matches if any of the inner selectors
+  // match. The inner array are simple selectors like `.foo` and `#bar` which
+  // must all match.
+  return dom5.predicates.OR(...parsed.map((simpleSelectors) => {
+    return dom5.predicates.AND(
+        ...simpleSelectors.map(simpleSelectorToPredicate));
+  }));
+}
+
+function simpleSelectorToPredicate(selector: cssWhat.Simple) {
+  switch (selector.type) {
+    case 'adjacent':
+    case 'child':
+    case 'descendant':
+    case 'parent':
+    case 'sibling':
+    case 'pseudo':
+      throw new Error(`Unsupported CSS operator: ${selector.type}`);
+    case 'attribute':
+      return attributeSelectorToPredicate(selector);
+    case 'tag':
+      return dom5.predicates.hasTagName(selector.name);
+    case 'universal':
+      return () => true;
+  }
+  const never: never = selector;
+  throw new Error(`Unexpected node type from css parser: ${never}`);
+}
+
+function attributeSelectorToPredicate(selector: cssWhat.Attribute):
+    dom5.Predicate {
+  switch (selector.action) {
+    case 'exists':
+      return dom5.predicates.hasAttr(selector.name);
+    case 'equals':
+      return dom5.predicates.hasAttrValue(selector.name, selector.value);
+    case 'start':
+      return (el) => {
+        const attrValue = dom5.getAttribute(el, selector.name);
+        return attrValue != null && attrValue.startsWith(selector.value);
+      };
+    case 'end':
+      return (el) => {
+        const attrValue = dom5.getAttribute(el, selector.name);
+        return attrValue != null && attrValue.endsWith(selector.value);
+      };
+    case 'element':
+      return dom5.predicates.hasSpaceSeparatedAttrValue(
+          selector.name, selector.value);
+    case 'any':
+      return (el) => {
+        const attrValue = dom5.getAttribute(el, selector.name);
+        return attrValue != null && attrValue.includes(selector.value);
+      };
+  }
+  const never: never = selector.action;
+  throw new Error(
+      `Unexpected type of attribute matcher from CSS parser ${never}`);
+}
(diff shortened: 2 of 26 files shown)


src/polymer/iron-flex-layout-classes.ts
```diff
@@ -0,0 +1,247 @@
...
+
+import stripIndent = require('strip-indent');
+
+import cssWhat = require('css-what');
+
+const p = dom5.predicates;
+
+const styleModules = [{
+  name: 'iron-flex',
+  selector: elementSelectorToPredicate('.layout.horizontal, .layout.vertical, .layout.inline, .layout.wrap, .layout.no-wrap, .layout.center, .layout.center-center, .layout.center-justified, .flex, .flex-auto, .flex-none')
```
rictic: please split to 80 cols. Maybe something like:

```javascript
'.layout.horizontal, .layout.vertical, .layout.inline, .layout.wrap,' + 
' .layout.no-wrap, .layout.center, .layout.center-center, .layout.center-justified, ' +
'.flex, .flex-auto, .flex-none'
```

src/polymer/iron-flex-layout-classes.ts
```diff
@@ -0,0 +1,247 @@
...
+    start: { line: tagRange.end.line, column: tagRange.end.column },
+    end: { line: tagRange.end.line, column: tagRange.end.column }
+  };
+}
+
+registry.register(new IronFlexLayoutClasses());
+
+// TODO(valdrin) move this in commons or something.
+/* ---- START copy-pasted from content-to-slot-usages. --- */
+function elementSelectorToPredicate(simpleSelector: string): dom5.Predicate {
```
rictic: This has merged in, so after rebase you should be able to move this into html/util

rictic (commented):
A couple early suggestions

src/polymer/iron-flex-layout-classes.ts
rictic: Also you can run `npm run format` to run the auto-formatter over the project. (or get it integrated into your editor on save, it's clang-format)

src/polymer/iron-flex-layout-classes.ts
```diff
@@ -0,0 +1,244 @@
...
+  {
+    name: 'iron-flex-factors',
+    selector: elementSelectorToPredicate(
+        '.flex-1, .flex-2, .flex-3, .flex-4, .flex-5, .flex-6, .flex-7, ' +
+        '.flex-8, .flex-9, .flex-10, .flex-11, .flex-12')
+  },
+  {
+    name: 'iron-positioning',
+    selector: elementSelectorToPredicate(
+        '.block, [hidden], .invisible, .relative, .fit, body.fullbleed, ' +
```
valdrinkoshi: should remove `[hidden]` as it's too generic

src/polymer/iron-flex-layout-classes.ts
```diff
@@ -0,0 +1,246 @@
...
+      const missingModules = getMissingStyleModules(templateContent);
+      if (!missingModules) {
+        continue;
+      }
+      // TODO(valdrin): update the warning location to be at the spot where
+      // the class is used.
+      const warning =
+          createWarning(parsedDocument, domModule.astNode, missingModules);
+      const styleNode = dom5.query(templateContent, p.hasTagName('style'));
+      if (!styleNode) {
```
valdrinkoshi: this could be simplified by just always adding a new style element, even if there are already other style elements in the template, since Polymer handles multiple styles anyways

CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Support warning about and automatically fixing element declarations that use `<content>` to use `<slot>` instead. This is a breaking change to an element,
 so it is not on by default.
   - It automatically adds the `old-content-selector` attribute to migrated `<slot>` elements, so uses can be automatically upgraded as well.
+- Support fixable warning about usage of deprecated `iron-flex-layout/classes/*.html` files via the new rule `iron-flex-layout-import`
```
rictic: Recommend a standard wording:

`Support warning about and fixing usage of…`

CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
...
+  - It automatically upgrades to import instead the `iron-flex-layout/iron-flex-layout-classes.html`
```
rictic: nit: 

```
to import `iron-flex-layout/iron-flex-layout-classes.html` instead.
```

src/polymer/iron-flex-layout-classes.ts
```diff
@@ -0,0 +1,262 @@
...
+class IronFlexLayoutClasses extends HtmlRule {
+  code = 'iron-flex-layout-classes';
+  description = stripIndent(`
+    Warns when iron-flex-layout classes are used without including the style modules.
+  `).trim();
+
+  async checkDocument(parsedDocument: ParsedHtmlDocument, document: Document) {
+    const warnings: FixableWarning[] = [];
+
+    this.convertDeclarations(parsedDocument, document, warnings);
```
rictic: I'd inline `convertDeclarations`. I did that in `content-to-slot-declarations` back when the usages and declarations passes were in a single rule rather than split out into two.

src/polymer/iron-flex-layout-classes.ts
```diff
@@ -0,0 +1,262 @@
...
+${indent}  <style is="custom-style" include="${missingModules}"></style>
+${indent}</custom-style>`,
+      range: sourceRangeForPrependContent(parsedDocument, body)
+    }];
+    warnings.push(warning);
+  }
+}
+
+function getMissingStyleModules(node: dom5.Node) {
+  const usedModules = styleModules.filter((m) => !!dom5.query(node, m.selector))
```
rictic: This does N passes over the document, one for each entry in `styleModules`. I'd recommend instead using dom5.nodeWalkAll and applying the predicate yourself. Maybe building a set of the modules that match.

src/polymer/iron-flex-layout-classes.ts
```diff
@@ -0,0 +1,262 @@
...
+${indent}<custom-style>
+${indent}  <style is="custom-style" include="${missingModules}"></style>
+${indent}</custom-style>`,
+      range: sourceRangeForPrependContent(parsedDocument, body)
+    }];
+    warnings.push(warning);
+  }
+}
+
+function getMissingStyleModules(node: dom5.Node) {
```
rictic: Consider also returning the first element that matches one of the styleModule predicates, so that you can place the warning on that node.

src/polymer/iron-flex-layout-classes.ts
```diff
@@ -0,0 +1,262 @@
...
+    if (!styleToEdit ||
+        styleModulesRegex.test(dom5.getAttribute(style, 'include')!)) {
+      styleToEdit = style;
+    }
+  }
+  // Fallback to style without include attribute.
+  return styleToEdit || dom5.query(node, p.hasTagName('style'));
+}
+
+function sourceRangeForAppendAttributeValue(
```
rictic: Consider taking the value to append and returning a Replacement

src/polymer/iron-flex-layout-classes.ts
```diff
@@ -0,0 +1,262 @@
...
+    parsedDocument: ParsedHtmlDocument, node: dom5.Node, attr: string) {
+  const tagRange = parsedDocument.sourceRangeForAttributeValue(node, attr)!;
+  return {
+    file: tagRange.file,
+    start: {line: tagRange.end.line, column: tagRange.end.column - 1},
```
rictic: What if the attribute's value isn't quoted?

src/polymer/iron-flex-layout-classes.ts
```diff
@@ -0,0 +1,262 @@
...
+    end: {line: tagRange.end.line, column: tagRange.end.column - 1}
+  };
+}
+
+function sourceRangeForAddAttribute(
+    parsedDocument: ParsedHtmlDocument, node: dom5.Node) {
```
rictic: Same two as above

src/polymer/iron-flex-layout-classes.ts
```diff
@@ -0,0 +1,262 @@
...
+  const tagRange = parsedDocument.sourceRangeForStartTag(node)!;
+  return {
+    file: tagRange.file,
+    start: {line: tagRange.end.line, column: tagRange.end.column - 1},
+    end: {line: tagRange.end.line, column: tagRange.end.column - 1}
+  };
+}
+
+function sourceRangeForPrependContent(
```
rictic: Same as above, recommend returning a Replacement here

src/polymer/iron-flex-layout-classes.ts
```diff
@@ -0,0 +1,262 @@
...
+    parsedDocument: ParsedHtmlDocument, node: dom5.Node) {
+  const tagRange = parsedDocument.sourceRangeForStartTag(node)!;
+  return {
+    file: tagRange.file,
+    start: {line: tagRange.end.line, column: tagRange.end.column},
+    end: {line: tagRange.end.line, column: tagRange.end.column}
+  };
+}
+
+registry.register(new IronFlexLayoutClasses());
```
rictic: I think this line should be at the bottom of the file.

src/polymer/iron-flex-layout-classes.ts
```diff
@@ -0,0 +1,262 @@
...
+
+// TODO(valdrin) move this in commons or something.
+/* ---- START copy-paste from move-style-into-template. ---- */
+function getIndentationInside(parentNode: dom5.Node) {
```
rictic: Move to util?

src/polymer/iron-flex-layout-import.ts
```diff
@@ -0,0 +1,74 @@
...
+          parsedDocument.sourceRangeForAttributeValue(imp, 'href')!;
+      const warning = new FixableWarning({
+        code: 'iron-flex-layout-import',
+        message: `${href} import is deprecated. ` +
+            `Replace with ${correctImport} import.`,
+        parsedDocument,
+        severity: Severity.WARNING,
+        sourceRange: hrefRange
+      });
+      warning.fix = [{replacementText: `"${correctImport}"`, range: hrefRange}];
```
rictic: Hm, this isn't an automatically safe fix. It's only safe when used with the rule above that adds style imports. Maybe we should start by just warning, not fixing?

src/polymer/iron-flex-layout-classes.ts
```diff
@@ -89,7 +89,18 @@ class IronFlexLayoutClasses extends HtmlRule {
       parsedDocument: ParsedHtmlDocument, document: Document,
       warnings: FixableWarning[]) {
     // Search in the dom-modules.
-    for (const domModule of document.getFeatures({kind: 'dom-module'})) {
+    for (const domModule of document.getFeatures({ kind: 'dom-module' })) {
+      const misplacedStyle = dom5.query(domModule.astNode, p.hasTagName('style'))
+      if (misplacedStyle) {
+        warnings.push(new FixableWarning({
```
rictic: I think we should just silently continue here. They'll already get a warning for style-into-template.

rictic (commented):
Great work!

src/polymer/iron-flex-layout-import.ts
valdrinkoshi: I'd keep the fix because this is a separate rule, not included in any collection. Also, I don't think there's the need of a warning as the user already gets a warning at runtime [by the `v1 iron-flex-layout/classes/*.html` files](https://github.com/PolymerElements/iron-flex-layout/blob/1.x/classes/iron-flex-layout.html#L14), and would get a EOENT error for v2 iron-flex-layout as these files don't exist. WDYT?

src/polymer/iron-flex-layout-classes.ts
valdrinkoshi: Good point, will just delete this method and rather replace the attribute value

src/polymer/iron-flex-layout-classes.ts
valdrinkoshi: Done

src/polymer/iron-flex-layout-classes.ts
valdrinkoshi: Done

src/polymer/iron-flex-layout-classes.ts
valdrinkoshi: Done

valdrinkoshi (commented):
Yay thanks for the review!
Re placing of the warning: currently we parse all the document and we find multiple matches, and we group them into 1 single fix. Is the suggestion to split these into individual fixes? The complicated bit would be to avoid adding a new style w/ include for each fix.
I guess I could have multiple warnings but only 1 with the fix...does that makes sense?

src/polymer/iron-flex-layout-classes.ts
valdrinkoshi: I've changed this to minimize the queries by first checking the used modules, and then avoid querying nodes if the module is already imported. I'm not sure how nodeWalkAll would save us from reducing the queries..

rictic:
Multiple warnings but only putting the fix on one… yeah that makes sense to me.

Or you could keep it at one warning and one fix, but place it at the first site where you notice an iron-flex-layout class.

src/polymer/iron-flex-layout-import.ts
rictic: +1

Mind adding a comment to this effect, as I think we should build out a way to express changes that aren't just automatically safe fixes of warnings. Then we could add this rule into the default collections.

src/polymer/iron-flex-layout-classes.ts
rictic: In our experience with the analyzer, code shaped like this:

```
for node in ast:
  for thing_to_do in things_to_do:
    thing_to_do(node)
```

is much faster than

```
for thing_to_do in things_to_do:
  for node in ast:
    thing_to_do(node)
```

This may be less true of HTML than javascript though, as the JS AST is more heterogenous, so it takes more effort to walk the tree.

rictic:
RE: nodeWalkAll vs multiple queries

In our experience with the analyzer, code shaped like this:

```
for node in ast:
  for thing_to_do in things_to_do:
    thing_to_do(node)
```

is much faster than

```
for thing_to_do in things_to_do:
  for node in ast:
    thing_to_do(node)
```

This may be less true of HTML than javascript though, as the JS AST is more heterogenous, so it takes more effort to walk the tree.

src/polymer/iron-flex-layout-classes.ts
```diff
@@ -0,0 +1,228 @@
...
+  return styleToEdit || dom5.query(node, p.hasTagName('style'));
+}
+
+function sourceRangeForAddAttribute(
+    parsedDocument: ParsedHtmlDocument, node: dom5.Node) {
+  const tagRange = parsedDocument.sourceRangeForStartTag(node)!;
+  return {
+    file: tagRange.file,
+    start: {line: tagRange.end.line, column: tagRange.end.column - 1},
+    end: {line: tagRange.end.line, column: tagRange.end.column - 1}
```
rictic: This -1 also makes me nervous for unquoted attributes. Mind adding a test for that?

src/polymer/iron-flex-layout-classes.ts
```diff
@@ -0,0 +1,228 @@
...
+    if (!styleToEdit ||
+        styleModulesRegex.test(dom5.getAttribute(style, 'include')!)) {
+      styleToEdit = style;
+    }
+  }
+  // Fallback to style without include attribute.
+  return styleToEdit || dom5.query(node, p.hasTagName('style'));
+}
+
+function sourceRangeForAddAttribute(
```
rictic: Recommend returning a Replacement here.

valdrinkoshi: Done

src/polymer/iron-flex-layout-classes.ts
valdrinkoshi: The range is for the start tag, so the -1 will go right before the closing >, e.g. in `<div>` it would go to `<div`

src/polymer/iron-flex-layout-classes.ts
valdrinkoshi: Done

src/polymer/iron-flex-layout-import.ts
valdrinkoshi: Agreed. Updated the description and warning, LMK if that works

valdrinkoshi (commented):
Feedback incorporated 👌 As DX, I think it's better to have multiple warnings rather than one, e.g. say we have:
```html
<div class="layout horizontal"></div>
<div class="fit"></div>
```
The rule would find the user needs to import `iron-flex` and `iron-positioning`, but would highlight the first div, resulting in an odd message:
``` 
<div class="layout horizontal"></div>
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

Style module is used but not imported:

  iron-flex iron-positioning

Import them in the template style include.
```
Instead, having 2 warnings (and 1 fix) would look better:

``` 
<div class="layout horizontal"></div>
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
iron-flex style module is used but not imported
Import it in the template style include.


<div class="fit"></div>
~~~~~~~~~~~~~~~~~~
iron-positioning style module is used but not imported
Import it in the template style include.
```

WDYT?

valdrinkoshi:
Also, made iron-flex-layout-import rule slightly more intelligent, it now detects if the good import is already there and avoids duplicate imports

src/html/util.ts
```diff
@@ -169,3 +171,95 @@ export function getIndentationInside(parentNode: dom5.Node) {
...
+  }
+  const never: never = selector.action;
+  throw new Error(
+      `Unexpected type of attribute matcher from CSS parser ${never}`);
+}
+
+export function removeTrailingWhitespace(
+    textNode: dom5.Node, parsedDocument: ParsedHtmlDocument) {
+  const prevText = dom5.getTextContent(textNode);
+  const match = prevText.match(/\n?[ \t]*$/);
```
valdrinkoshi, line 250: Note: slight change on the regex from `/\n?[ \t]+$/` to `/\n?[ \t]*$/` aka it handles elements with no spaces - added a test for this

src/polymer/iron-flex-layout-import.ts
```diff
@@ -59,8 +59,9 @@ class IronFlexLayoutImport extends HtmlRule {
           parsedDocument.sourceRangeForAttributeValue(imp, 'href')!;
       const warning = new FixableWarning({
         code: 'iron-flex-layout-import',
-        message: `${href} import is deprecated. ` +
-            `Replace with ${correctImport} import.`,
+        message: `${href} import is deprecated in iron-flex-layout v1, and not shipped in iron-flex-layout v2.x.
+Replace it with ${correctImport} import.
+Run \`iron-flex-layout-classes\` to include the required style modules.`,
```
rictic: Recommend the wording:

Run the lint rule  \`iron-flex-layout-classes\` with `--fix` to include...

src/polymer/iron-flex-layout-classes.ts
```diff
@@ -158,31 +159,44 @@ ${indent}</custom-style>`,
...
-    includes += ' ' + dom5.getAttribute(style, 'include')!;
+  const modules = {};
+  dom5.nodeWalkAll(rootNode, (node: dom5.Node) => {
+    if (dom5.isElement(node)) {
+      if (isStyleInclude(node)) {
+        includes += ' ' + dom5.getAttribute(node, 'include')!;
+      } else {
+        styleModules.forEach((m) => {
+          if (!modules[m.module] && m.selector(node)) {
+            modules[m.module] = node;
```
rictic: Interesting. Did you consider making modules map to an array of nodes rather than a single one?

On the one hand, it could produce a flood of warnings. On the other hand, that's how people expect linters to work, and in the IDE use case it's much nicer to see all of the places where the classes are used incorrectly.

rictic (approved):
Really great work! A couple minor suggestions, but this is good to go.

src/polymer/iron-flex-layout-import.ts
valdrinkoshi: Done 👍

src/polymer/iron-flex-layout-classes.ts
valdrinkoshi: Updated to warn of missing style module include on every node using it 👌

valdrinkoshi:
@rictic incorporated last feedback, PTAL.

I've been facing few cases where the element defines and uses the `flex` class, and the linter rule erroneously warns about `iron-flex` not being imported (this is the limitation 1) described in the PR).
I'm probably overthinking this, but I'm afraid this might be annoying to users, who would have to choose to either change the class name not to get the warning (not ideal), or even ignore the warnings (worse - defeats the purpose of the warning).
I'm not sure we can improve enough this linter rule to handle all the cases...WDYT?

valdrinkoshi:
Uh oh! The rule doesn't work when a dom-module has a `<template is="dom-if">`! Should fix it

rictic:
The user can always exclude certain rules from running. If they're defining css classes with the same names as the iron-layout classes then they're gonna get some false positives and they should probably not use this rule.

src/polymer/iron-flex-layout-classes.ts
```diff
@@ -214,6 +198,37 @@ Import it in the template style include.`,
   return missingModules.trim();
 }
 
+function searchUsedModulesAndIncludes(
```
rictic: Using Objects as maps is an anti-pattern. Recommend `Map<string, dom5.Node[]>` instead for `modules`

src/polymer/iron-flex-layout-classes.ts
```diff
@@ -162,25 +154,25 @@ ${indent}<style include="${modules}"></style>`)];
...
+    const missingModules =
+        getMissingStyleModules(parsedDocument, parsedDocument.ast, warnings);
+    if (!missingModules) {
       return warnings;
     }
-    // TODO(valdrin): update the warning location to be at the spot where
-    // the class is used.
-    const warning = createWarning(parsedDocument, missingModules);
-    const modules = missingModules.map((m) => m.module).join(' ');
+    const warning = warnings[warnings.length - 1];
```
rictic: Why put this on the last warning? I'd recommend putting it on the first warning instead.

rictic (approved):

rictic:
Arg, github you're so dumb about code review. I commented on lines that changed, but not substantively.

PR #117, opened 2017-10-31
Update to typescript v2.6

rictic:

Just unused code. This may not compile until the analyzer has been updated.

---
This change is [Reviewable](https://reviewable.io/reviews/polymer/polymer-linter/117)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: node_js
 node_js:
   - "6"
+  - "8"
   - "node"
 script:
   - npm test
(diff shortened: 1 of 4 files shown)


valdrinkoshi (changes requested):
could you also update `src/test/util.ts`? it currently errors with (18,11): error TS6133: '_printer' is declared but its value is never read.

valdrinkoshi:
Also, `src/polymer/call-super-in-callbacks.ts` has issues with types in the `estraverse.traverse(` calls: Two different types with this name exist, but they are unrelated (?!)

rictic:
The estraverse error is due to `npm link`. An upcoming version of typescript will do this smarter I think (hope)

rictic:
Fixed the `_printer` error and updated to new analyzer.

valdrinkoshi (approved):

PR #122, opened 2017-11-10
Add actions to warnings.

rictic:
This represents more general actions that could be applied to address warnings. The first action kind is an edit, which is like a fix with caveats, and which needs an indication of specific user intent to apply.

This change also adds content-to-slot-declarations as a default rule, using edit actions to delineate changes which are safe from those that aren't.

---
This change is [Reviewable](https://reviewable.io/reviews/polymer/polymer-linter/122)

### DIFF
--- a/src/polymer/content-to-slot-declarations.ts
+++ b/src/polymer/content-to-slot-declarations.ts
@@ -19,9 +19,8 @@ import {Document, ParsedHtmlDocument, Severity} from 'polymer-analyzer';
 
 import {HtmlRule} from '../html/rule';
 import {registry} from '../registry';
+import {stripIndentation} from '../util';
 import {Edit, FixableWarning} from '../warning';
-
-import stripIndent = require('strip-indent');
 
 
 const p = dom5.predicates;
@@ -32,9 +31,16 @@ const p = dom5.predicates;
 
 class ContentToSlotDeclarations extends HtmlRule {
   code = 'content-to-slot-declarations';
-  description = stripIndent(`
+  description = stripIndentation(`
       Warns when using <content> instead of Shadow Dom v1's <slot> element.
-  `).trim();
+
+      This warning is automatically fixable, and also supports an edit action
+      to convert:
+          <content select=".foo"></content>
+
+      To:
+          <slot name="foo" old-content-selector=".foo"></slot>
+  `);
 
   async checkDocument(parsedDocument: ParsedHtmlDocument, document: Document) {
     const warnings: FixableWarning[] = [];
@@ -62,9 +68,9 @@ class ContentToSlotDeclarations extends HtmlRule {
         });
         const result = getSerializedSlotElement(contentElement, slotNames);
         if (result !== undefined) {
-          const {slotElementText, isSafe} = result;
-          const slotElementStartTag =
-              slotElementText.slice(0, -7); /* cut </slot> off the end */
+          const {slotElementStartTagText, isSafe} = result;
+          const slotElementStartTag = slotElementStartTagText.slice(
+              0, -7); /* cut </slot> off the end */
           const edit: Edit = [
             {
               replacementText: slotElementStartTag,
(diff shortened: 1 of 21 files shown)


valdrinkoshi:
The implementation LGTM, but I cannot test it locally - I have polymer-analyzer and polymer-cli correctly linked, but I don't get the "what should be done?" prompt. Should I use a particular branch for polymer-analyzer or polymer-cli?

rictic:
Yeah, check out the `lint-edit-actions` branch of polymer-cli, which builds on top of this

rictic:
Opened for review here: https://github.com/Polymer/polymer-cli/pull/924

valdrinkoshi:
Cool, thanks! I do see the prompt, but I noticed that running the rule `content-to-slot-declarations --fix` on this file would convert this:
```html
<dom-module id="foo-bar">
  <template>
    <content select=".foo">
      Fallback content here
    </content>
    <content></content>
  </template>
  <script>
    Polymer({is: 'foo-bar'});
  </script>
</dom-module>
``` 
to this (notice `<content></content>` was not converted):
```html
<dom-module id="foo-bar">
  <template>
    <slot name="foo" old-content-selector=".foo">
      Fallback content here
    </slot>
    <content></content>
  </template>
  <script>
    Polymer({is: 'foo-bar'});
  </script>
</dom-module>
```
Even if I choose to not apply the unsafe edits, it should apply the safe ones, and that doesn't seem to be the case

src/polymer/content-to-slot-declarations.ts
```diff
@@ -75,6 +75,23 @@ class ContentToSlotDeclarations extends HtmlRule {
...
+              kind: 'edit',
+              code: 'content-with-select',
+              description: stripIndent(`
+                Convert this <content> to a <slot>.
+
+                This changes the API of this element because the \`select\`
+                will become a slot name. Use the content-to-slot-usages lint
+                pass to convert usages of the element to conform to the new
+                API.
+            `).trim(),
```
justinfagnani: it'd be nice if this was built into stripIndent, and as a template tag.

rictic:
Ah, that was a bug in the CLI. Pushed a commit to fix that.

src/polymer/content-to-slot-declarations.ts
```diff
@@ -75,6 +75,23 @@ class ContentToSlotDeclarations extends HtmlRule {
...
+          if (isSafe) {
+            warning.fix = edit;
+          } else {
+            warning.actions = [{
+              kind: 'edit',
+              code: 'content-with-select',
+              description: stripIndent(`
+                Convert this <content> to a <slot>.
+
+                This changes the API of this element because the \`select\`
```
justinfagnani: "the select"? Maybe "the value of the select attribute"?

src/polymer/content-to-slot-declarations.ts
```diff
@@ -60,12 +60,12 @@ class ContentToSlotDeclarations extends HtmlRule {
           severity: Severity.WARNING,
           sourceRange: parsedDocument.sourceRangeForStartTag(contentElement)!
         });
-        const slotElementText =
-            getSerializedSlotElement(contentElement, slotNames);
-        if (slotElementText != null) {
+        const result = getSerializedSlotElement(contentElement, slotNames);
+        if (result !== undefined) {
+          const {slotElementText, isSafe} = result;
           const slotElementStartTag =
```
justinfagnani: Is this assuming there's no fallback content?

rictic: No, and we have tests for that. Renamed for clarity.

src/polymer/content-to-slot-declarations.ts
rictic: +1

src/polymer/content-to-slot-declarations.ts
rictic: Added a new util method `stripIndentation` to do both. I didn't make it a template tag as I didn't want to deal with the hassle of interpolation. Should we prefer functions on a single string to be template tags?

valdrinkoshi (approved):
👌

src/polymer/content-to-slot-declarations.ts
justinfagnani: I think it's just a bit nicer ergonomically, but then again it doesn't compose with other tags very easily, so 🤷‍♂️

PR #129, opened 2017-11-16
iron-flex-layout-classes handles class$ attribute (fix 126)

### LINKED ISSUE #126: iron-flex-layout-classes doesn't handle class$ attribute
Classes used in class bindings like `<div class$="layout horizontal [[computedClass]]">` could be handled, but currently are not.

valdrinkoshi:

Fixes #126

---
This change is [Reviewable](https://reviewable.io/reviews/polymer/polymer-linter/129)


src/html/util.ts
```diff
@@ -173,27 +173,34 @@ export function getIndentationInside(parentNode: dom5.Node) {
...
  * between elements like `.foo .bar` or `.foo > .bar`.
+ *
+ * Pass `attributeAliases` to apply selectors in non-standard attributes, e.g.
+ * <div class$="foo bar"></div>
  */
-export function elementSelectorToPredicate(simpleSelector: string):
-    dom5.Predicate {
+export function elementSelectorToPredicate(
+    simpleSelector: string,
+    attributeAliases: AttributeAliases = {}): dom5.Predicate {
```
rictic: Hm, there are other cases where we'd like this behavior, and the attributeAliases syntax requires more knowledge from the caller.

What we want to say here is that if we're inside a Polymer template databinding context, then we want attribute matching to be slightly different. So if we had a boolean that indicated that we're in a polymer template, then for some of the attribute selectors (maybe `exists`, `element`, and `any`?) we would match on both the attribute name and the attribute name with a `$` suffix

valdrinkoshi: Updated to append $ for polymer templates. I did it for all actions, as we might want to have selectors like `[src^="//www.lorempixel.com/"]`, which corresponds to the action `start`, and should return nodes that are bound like this:
```html
<img src$="//www.lorempixel.com/[[dimension]]/[[dimension]]">
```

src/html/util.ts
```diff
@@ -186,20 +186,20 @@ export interface AttributeAliases { [s: string]: string; }
  */
```
rictic, line 189: Comment needs updating

rictic (approved):

src/html/util.ts
valdrinkoshi: touche! on it

PR #132, opened 2017-11-16
Return the analysis from the lint pass too.

rictic:

---
This change is [Reviewable](https://reviewable.io/reviews/polymer/polymer-linter/132)


rictic:
This is mostly useful for makeParseLoader, when called from the cli or the ide

valdrinkoshi (commented):
Not clear why is this needed for, my understanding is that we now add an `analysis` property to an array. Is this needed internally only (for the tests) or elsewhere too?

rictic:
It's also useful in the CLI and IDE which call the same makeParseLoader function

valdrinkoshi (approved):
👍

PR #152, opened 2018-01-11
Add iron-form-v1-to-v2 rule

valdrinkoshi:

---
This change is [Reviewable](https://reviewable.io/reviews/polymer/polymer-linter/152)

### DIFF
--- a/src/polymer/elements/iron-form-v1-to-v2.ts
+++ b/src/polymer/elements/iron-form-v1-to-v2.ts
@@ -27,25 +27,25 @@ const isIronFormV1 = p.AND(
     p.hasTagName('form'),
     p.hasAttrValue('is', 'iron-form'));
 
-const propertiesToDelete = [
+const propertiesToDelete: ReadonlySet<string> = new Set([
   'is',
   'disable-native-validation-ui',
   'disable-native-validation-ui$',
   'request',
   'request$',
-];
-const propertiesToMove = [
+]);
+const propertiesToMove: ReadonlySet<string> = new Set([
   'id',
   'id$',
   'headers',
   'headers$',
   'with-credentials',
   'with-credentials$'
-];
-const propertiesToRename: {[key: string]: string;} = {
-  'content-type': 'enctype',
-  'content-type$': 'enctype$',
-};
+]);
+const propertiesToRename: ReadonlyMap<string, string> = new Map([
+  ['content-type', 'enctype'],
+  ['content-type$', 'enctype$'],
+]);
 
 class IronFormV1ToV2 extends HtmlRule {
   code = 'iron-form-v1-to-v2';
@@ -82,11 +82,11 @@ class IronFormV1ToV2 extends HtmlRule {
       form.attrs.forEach((attr) => {
         // All `iron-form-*` events and `propertiesToMove` go on <iron-form>.
         if ((attr.name.indexOf('on-iron-form-') === 0 ||
-             propertiesToMove.indexOf(attr.name) >= 0)) {
+             propertiesToMove.has(attr.name))) {
           ironFormAttrs +=
               ` ${attr.name}${attr.value ? `="${attr.value}"` : ''}`;
-        } else if (propertiesToDelete.indexOf(attr.name) === -1) {
-          const attrName = propertiesToRename[attr.name] || attr.name;
+        } else if (!propertiesToDelete.has(attr.name)) {
+          const attrName = propertiesToRename.get(attr.name) || attr.name;
           formAttrs += ` ${attrName}${attr.value ? `="${attr.value}"` : ''}`;
         }
       });
--- a/src/polymer/elements/iron-form-v1-to-v2.ts
+++ b/src/polymer/elements/iron-form-v1-to-v2.ts
@@ -13,7 +13,7 @@
  */
 
 import * as dom5 from 'dom5';
-import {Document, ParsedHtmlDocument, Severity, Warning} from 'polymer-analyzer';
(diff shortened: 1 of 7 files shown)


stramel:
@valdrinkoshi any chance you are going to be able to revisit this?

valdrinkoshi:
@stramel I'm hesitant on the `--fix` part as the new iron-form behaves quite differently compared to the 1.x..I should probably document more this in the warning. I'll get to this soon, it's on my radar 👌

valdrinkoshi:
Comparing the properties exposed in 1.x to 2.x:
iron-form 1.x has these properties https://www.webcomponents.org/element/PolymerElements/iron-form/v1.1.6/elements/iron-form
iron-form 2.x has these properties https://www.webcomponents.org/element/PolymerElements/iron-form/elements/iron-form

Based on that, `--fix` should do this:
- wrap `<form>` into an `<iron-form>`
- move `on-iron-form-*` events from `<form>` to `<iron-form>`
- move `id, withCredentials, headers` from `<form>` to `<iron-form>`
- rename `contentType` to `enctype` (keep it on `<form>`)
- remove `disableNativeValidationUi, request` properties from `<form>` (iron-form 2.x doesn't expose these)

src/rule.ts
```diff
@@ -29,8 +29,8 @@ export abstract class Rule {
    */
   abstract readonly description: string;
 
-  protected readonly cache =
-      new WeakMap<Document, Promise<ReadonlyArray<Warning>>>();
+  protected readonly cache: WeakMap<Document, Promise<ReadonlyArray<Warning>>> =
```
rictic: why this change?

src/polymer/elements/iron-form-v1-to-v2.ts
```diff
@@ -0,0 +1,135 @@
...
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import * as dom5 from 'dom5';
```
rictic: Should switch to the new dom5 API:

`import * as dom5 from 'dom5/lib/index-next';`

src/polymer/elements/iron-form-v1-to-v2.ts
```diff
@@ -0,0 +1,135 @@
...
+              <input type="text">
+              <input type="submit">
+            </form>
+          </iron-form>
+  `);
+
+  async checkDocument(parsedDocument: ParsedHtmlDocument, _document: Document) {
+    const warnings: Warning[] = [];
+    const forms = dom5.queryAll(
+        parsedDocument.ast, isIronFormV1, [], dom5.childNodesIncludeTemplate);
```
rictic: Changes in dom5.queryAll:

* The third parameter is dropped entirely
* It returns an iterable rather than an array
* Should be a bit faster!

src/polymer/elements/iron-form-v1-to-v2.ts
```diff
@@ -0,0 +1,135 @@
...
+    for (const form of forms) {
+      // Collect attributes/events for the iron-form and form.
+      let ironFormAttrs = '', formAttrs = '';
+      form.attrs.forEach((attr) => {
+        // All `iron-form-*` events and `propertiesToMove` go on <iron-form>.
+        if ((attr.name.indexOf('on-iron-form-') === 0 ||
+             propertiesToMove.indexOf(attr.name) >= 0)) {
+          ironFormAttrs +=
+              ` ${attr.name}${attr.value ? `="${attr.value}"` : ''}`;
```
rictic: This is battle tested, so it's fine, but for future passes I think we should try to use dom5 APIs as much as possible rather than building HTML with string manipulation. We need more convenience methods for doing this.

It's not a security thing, as this is transforming from trusted code to trusted code, it's just less error prone.

src/polymer/elements/iron-form-v1-to-v2.ts
```diff
@@ -0,0 +1,135 @@
...
+    p.hasAttrValue('is', 'iron-form'));
+
+const propertiesToDelete = [
+  'is',
+  'disable-native-validation-ui',
+  'disable-native-validation-ui$',
+  'request',
+  'request$',
+];
+const propertiesToMove = [
```
rictic: Recommend making these two `ReadonlySet<string>`

src/polymer/elements/iron-form-v1-to-v2.ts
```diff
@@ -0,0 +1,135 @@
...
+  'id',
+  'id$',
+  'headers',
+  'headers$',
+  'with-credentials',
+  'with-credentials$'
+];
+const propertiesToRename: {[key: string]: string;} = {
```
rictic: Recommend making this a ReadonlyMap<string, string>

rictic (changes requested):
(using the Request Changes review response, despite the big ugly red X because it's more legible to tooling)

src/rule.ts
valdrinkoshi: I think I was facing some type infer kind of issues when trying it with analyzer + cli. I reverted this for now, will open a separate PR if needed

valdrinkoshi (commented):
Feedback incorporated 👌

rictic (approved):

rictic:
Awesome work on this PR and the mass migration internally using it!

PR #154, opened 2018-02-04
add Polymer.dom calls rule

### LINKED ISSUE #143: Polymer.dom warnings
Could easily warn about usages of `Polymer.dom` and potentially fix usages.

**Examples:**
- `Polymer.dom(event).localTarget` to `event.target`
- `Polymer.dom(event).path` to `event.composedPath()`
- `Polymer.dom(event).rootTarget` to `event.composedPath()[0]`

43081j:
Fixes #143.

---

I was having a look around the linter and this one seemed simple enough to implement. 

Couple of questions:
* Is `Polymer.dom` actually deprecated? if so, should i update the messages to say so?
* I put this in the polymer 2 ruleset as 1.x and hybrid elements still depend on it, right?
* If we want to implement fixes/replacements, we need some way of turning an expression into a string and i don't think the analyser exposes our util method for that (e.g. convert the ast of `some.variable.here` to the literal string `"some.variable.here"`, and more complex cases too)

---
This change is [Reviewable](https://reviewable.io/reviews/polymer/polymer-linter/154)


src/polymer/dom-calls-to-native.ts
```diff
@@ -0,0 +1,121 @@
...
+      babel.isIdentifier(expr.object.callee.object) &&
+      babel.isIdentifier(expr.object.callee.property) &&
+      expr.object.callee.object.name === 'Polymer' &&
+      expr.object.callee.property.name === 'dom';
+  }
+}
+
+registry.register(new DomCallsToNative());
+
+function getParsedDocumentContaining(
```
43081j: this thing needs to go away... we have a method now in the analyzer (somewhere inside `Analysis`). but im not sure how we get at it...

stramel:
@43081j my hero 👍🏼

src/polymer/dom-calls-to-native.ts
```diff
@@ -0,0 +1,118 @@
+/**
+ * @license
+ * Copyright (c) 2017 The Polymer Project Authors. All rights reserved.
```
stramel: Technically I think this should be the year this file was created.

src/test/polymer/dom-calls-to-native.ts
```diff
@@ -0,0 +1,69 @@
+/**
+ * @license
+ * Copyright (c) 2017 The Polymer Project Authors. All rights reserved.
```
stramel: Same here

src/polymer/dom-calls-to-native.ts
```diff
@@ -0,0 +1,118 @@
...
+
+import babelTraverse from 'babel-traverse';
+import * as babel from 'babel-types';
+import {Document, Severity, Warning, SourceRange, ParsedDocument, isPositionInsideRange} from 'polymer-analyzer';
+
+import {registry} from '../registry';
+import {Rule} from '../rule';
+import {stripWhitespace} from '../util';
+
+import {stripIndentation} from '../util';
```
stramel: Can you consolidate these imports?

43081j:
@stramel i've gone ahead and sorted those things

& also ran the formatter on it to clean up some whitespace etc.

src/polymer/dom-calls-to-native.ts
```diff
@@ -0,0 +1,117 @@
...
+              getParsedDocumentContaining(doc.sourceRange, document);
+
+          if (!containingDoc) {
+            return;
+          }
+
+          const sourceRange = containingDoc.sourceRangeForNode(path.node)!;
+
+          if (this._isPolymerDomCall(path.node) &&
+              this._replacements.has(name)) {
```
rictic: Let's move these checks up above the work to get the containing doc, as they're very fast checks, and they don't allocate, whereas containingDoc and sourceRangeForNode are a bit more expensive.

Early exiting with `return` if these don't match is nice too.

src/polymer/dom-calls-to-native.ts
```diff
@@ -0,0 +1,117 @@
...
+          const containingDoc =
+              getParsedDocumentContaining(doc.sourceRange, document);
+
+          if (!containingDoc) {
+            return;
+          }
+
+          const sourceRange = containingDoc.sourceRangeForNode(path.node)!;
+
+          if (this._isPolymerDomCall(path.node) &&
```
rictic: I think it'd make a bit more sense if we passed in `path.node.object` to `_isPolymerDomCall`

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 <!-- ## Unreleased -->
```
rictic: Uncomment out the `## Unreleased` section please

rictic: Uncomment out the `## Unreleased` section please

rictic (commented):
Very cool! This will be a huge help for people migrating!

A few small suggestions and this is good to land

rictic:
Ignore the node travis failure, that's node v6 getting TypeScript 2.7 which is a bit more strict on us

43081j:
@rictic i've updated as per your comments.

do you know if i can get the containing doc function from the analyzer somehow? it feels a bit hacky having to pretty much re-define it here

src/polymer/dom-calls-to-native.ts
```diff
@@ -0,0 +1,119 @@
...
+
+          const containingDoc =
+              getParsedDocumentContaining(doc.sourceRange, document);
+
+          if (!containingDoc) {
+            return;
+          }
+
+          const sourceRange = containingDoc.sourceRangeForNode(path.node)!;
+          const replacement = this._replacements.get(name)!;
```
rictic: The ! is a code smell. Consider:

```typescript
const replacement = this._replacements.get(name);

 if (!this._isPolymerDomCall(path.node.object) || replacement === undefined) {
  return;
}
```

and similarly early exiting if sourceRange is undefined (I know there's no current way for that source range to be undefined, but good type checking has caught a lot of other things that didn't seem likely to change at the time that I prefer to just lean heavily on the type checker wherever we can)

src/polymer/dom-calls-to-native.ts
rictic: Yeah, we need to pass down the `analysis` to `checkDocument`. Mind putting a TODO and I'll tackle that in a followup PR?

rictic (commented):
Just two more super quick things, if you don't mind!

43081j:
here you go

rictic (approved):

rictic:
Awesome!

PR #157, opened 2018-02-13
Fix: Better guess for label slot

stramel:

I noticed that `paper-input-container`s with `label` are getting marked with `slot="input"`. The previous selector was for "input" and "label" (`:not([add-on]):not([prefix]):not([suffix])`)

---
This change is [Reviewable](https://reviewable.io/reviews/polymer/polymer-linter/157)


rictic:
Ah tricky, yeah in cases like these where one `<content>` has become two `<slot>`s this is probably the best we can do.

rictic (approved):

stramel:
Figured it was better than previously marking it as `slot="input"` lol Thanks for the quick merge!

PR #159, opened 2018-02-14
Fix: build issues

stramel:
/cc @rictic

---
This change is [Reviewable](https://reviewable.io/reviews/polymer/polymer-linter/159)

### DIFF
--- a/src/rule.ts
+++ b/src/rule.ts
@@ -12,7 +12,9 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
-import {Document, Warning} from 'polymer-analyzer';
+// TODO(stramel): Remove on drop of Node 6
+// @ts-ignore
+import {Document, Warning, ParsedDocument} from 'polymer-analyzer';
 
 /**
  * A lint rule. Can take a package and find Warnings.
--- a/src/rule.ts
+++ b/src/rule.ts
@@ -12,8 +12,6 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
-// TODO(stramel): Remove on drop of Node 6
-// @ts-ignore
 import {Document, Warning, ParsedDocument} from 'polymer-analyzer';
 
 /**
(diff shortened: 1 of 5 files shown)


src/rule.ts
```diff
@@ -12,7 +12,9 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
-import {Document, Warning} from 'polymer-analyzer';
+// TODO(stramel): Remove on drop of Node 6
+// @ts-ignore
```
rictic: What's up with this annotation?

rictic: Ah, because TypeScript can't name ParsedDocument in the emitted declarations. This is a TypeScript 2.7 thing, not a node v6 thing, it's just that node v6 doesn't understand package locks, so it pulls in typescript latest.

I fixed this locally by importing ParsedDocument and doing this at the bottom of the file. It's ugly but it works:

```typescript
+// Need to have ParsedDocument in exported types for some reason.
+if (false as true) {
+  console.log(null as ParsedDocument | null);
+}
```

stramel: Ah, I figured it was something like that.

`@ts-ignore` ignores the next line

Should I push something like this to it?

```js
// Need to have ParsedDocument in exported types for some reason.
const t: null|ParsedDocument = null;
t;
```

or would you prefer just removing this "fix"

rictic: that works too.

stramel: @rictic Would you prefer the usage of `ParsedDocument`, the `@ts-ignore` comment directive, or setting the package.json to use `~2.6.2` instead of `^`?

stramel:
Timeout on Node 8 during integration tests

src/rule.ts
rictic: Usage of ParsedDocument, definitely. We want to be on 2.7 anyways (it caught a legit typo in a test! `:` instead of `=`), and we don't want to be in the business of ignoring compiler warnings.

rictic (approved):

rictic:
Travis has been timing out more often lately. There's an open bug on analyzer where we're looking into it, but this PR is ready to land.

rictic:
Thanks!

stramel:
You're welcome!

PR #160, opened 2018-02-14
New: Warn of `::content` usages

stramel:

Wasn't too fond of the naming of it. 

**Side Note:** I feel like the `content` -> `slot` rule should be under HTML instead of Polymer.

It would be cool to allow collections to include other collections, like a WebComponents v0 to v1 collection that could be included in some of the Polymer collections.

/cc @rictic

---
This change is [Reviewable](https://reviewable.io/reviews/polymer/polymer-linter/160)


src/css/content-to-slotted-usages.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  async checkDocument(parsedDocument: ParsedCssDocument, _document: Document) {
+    const warnings: Warning[] = [];
+
+    for (const node of parsedDocument) {
+      if (node.type === shady.nodeType.ruleset) {
+        const deprecatedRegex = /::content/;
+        const match = node.selector.match(deprecatedRegex);
+
+        if (match) {
```
rictic: Prefer more specific comparisons in conditionals, so:

```typescript
if (match !== undefined) {
  //
}
```

This is faster at runtime, and the type system will help more with them, which keeps us robust when we refactoring.

src/css/content-to-slotted-usages.ts
```diff
@@ -0,0 +1,84 @@
...
+              }
+            ];
+          }
+
+          warnings.push(new Warning({
+            code: 'content-to-slotted-usages',
+            severity: Severity.WARNING, parsedDocument, sourceRange,
+            message:
+                'The ::content pseudo-element has been deprecated in favor of ' +
+                'the ::shadow psuedo-element in WebComponents v1.',
```
rictic: ::shadow -> ::slotted

WebComponts v1 -> ShadowDOM v1

src/css/content-to-slotted-usages.ts
```diff
@@ -0,0 +1,84 @@
...
+
+import {Document, ParsedCssDocument, Replacement, Severity, Warning} from 'polymer-analyzer';
+import * as shady from 'shady-css-parser';
+
+import {CssRule} from '../css/rule';
+import {registry} from '../registry';
+import {stripIndentation} from '../util';
+
+class ContentToSlottedUsages extends CssRule {
+  code = 'content-to-slotted-usages';
```
rictic: How about `'content-selector-to-slotted'`. 

The `usages` isn't necessary, because there isn't another form (we have two passes for `<content>`, one that gets declarations of `<content>` and the other that gets at places where elements expected to be distributed into a `<content>`).

test/content-to-slotted-usages/before-fixes.html
```diff
@@ -0,0 +1,18 @@
...
+  <style>
+    :host ::content > [top-item] {
+      color: #F00;
+    }
+    ::content > top-item {
+      color: #0F0;
+    }
+    ::content > * {
+      color: #00F;
+    }
```
rictic: I'd like to see a couple examples of `::content` usages here that can't be fixed.

rictic:
Also +1 to moving a number of the rules out from under `Polymer`. A number are there because we didn't have a general mechanism for a lint pass to get the template of an element, but we're adding that as part of Polymer 3 support.

rictic:
Looks like the integration tests are broken, we should disable them for now, as they haven't been maintained in a while.

src/css/content-to-slotted-usages.ts
stramel: doh! 😄

test/content-to-slotted-usages/before-fixes.html
stramel: The test harness for comparing before and after expects no "incompatible edits"

`assert.deepEqual(incompatibleEdits, []);` Should I modify this to allow for "incompatible edits"?

rictic: An incompatible edit is one where two replacements (in one edit, or across multiple edits) overlap such that both of them can't be applied.

It's ok if there are warnings without fixes.

src/css/content-selector-to-slotted.ts
```diff
@@ -66,11 +66,11 @@ class ContentToSlottedUsages extends CssRule {
           }
 
           warnings.push(new Warning({
-            code: 'content-to-slotted-usages',
+            code: 'content-selector-to-slotted',
             severity: Severity.WARNING, parsedDocument, sourceRange,
             message:
                 'The ::content pseudo-element has been deprecated in favor of ' +
-                'the ::shadow psuedo-element in WebComponents v1.',
+                'the ::slotted psuedo-element in WebComponents v1.',
```
rictic: Last thing: WebComponents v1 -> ShadowDOM v1

rictic:
Arg, github's comment collapsing algorithm. Copy-pasta:

One last thing is: WebComponents v1 -> ShadowDOM v1

src/css/content-selector-to-slotted.ts
stramel: Oh, sorry, I missed that.

stramel:
Also, seems like appveyor is never actually running?

rictic:
Yeah, looks like appveyor isn't clearing: https://ci.appveyor.com/project/rictic/polymer-linter/history

Apparently this isn't an uncommon problem: https://help.appveyor.com/search?q=queued&recent=1&t=d&utf8=%E2%9C%93 and the solution is to just ask them to manually fix it. Filed as https://help.appveyor.com/discussions/problems/12123-all-builds-queued

rictic:
Thanks for the PR!

PR #162, opened 2018-02-14
 New: Add safe fixes for dom-module

stramel:

/cc @rictic

---
This change is [Reviewable](https://reviewable.io/reviews/polymer/polymer-linter/162)


src/test/html/dom-module-name-or-is_test.ts
```diff
@@ -18,11 +18,12 @@ import {Analyzer} from 'polymer-analyzer';
 
 import {Linter} from '../../linter';
 import {registry} from '../../registry';
-import {WarningPrettyPrinter} from '../util';
+import {assertExpectedFixes, WarningPrettyPrinter} from '../util';
 
 const fixtures_dir = path.join(__dirname, '..', '..', '..', 'test');
+const ruleId = 'dom-module-name-or-is';
```
rictic: Hm, looks like we've got an inconsistency in the name of the file vs the id of the rule. Mind moving files around to standardize on `dom-module-name-or-is`?

Would involve using ruleId in the getRules call in setup, and renaming the ts file and test directory

src/html/dom-module-name-or-is.ts
```diff
@@ -49,18 +49,32 @@ class DomModuleNameOrIs extends HtmlRule {
...
     for (const domModule of badModules) {
+      const isFixable =
+          !(dom5.getAttribute(domModule, 'is') !== null &&
+            dom5.getAttribute(domModule, 'name') !== null);
       for (const badAttr of ['is', 'name']) {
         const attr = dom5.getAttribute(domModule, badAttr);
-        if (attr != null) {
+        if (attr !== null) {
+          const sourceRange =
+              document.sourceRangeForAttributeName(domModule, badAttr)!;
```
rictic: Not your bug but I just noticed it. We should check to see if sourceRange is undefined, and `continue` if so.

src/html/dom-module-name-or-is.ts
```diff
@@ -49,18 +49,32 @@ class DomModuleNameOrIs extends HtmlRule {
         p.hasTagName('dom-module'), p.OR(p.hasAttr('is'), p.hasAttr('name')));
     const badModules = dom5.queryAll(document.ast, badModule);
     for (const domModule of badModules) {
+      const isFixable =
+          !(dom5.getAttribute(domModule, 'is') !== null &&
+            dom5.getAttribute(domModule, 'name') !== null);
       for (const badAttr of ['is', 'name']) {
         const attr = dom5.getAttribute(domModule, badAttr);
-        if (attr != null) {
+        if (attr !== null) {
```
rictic: Not your bug, but prefer to check and continue rather than adding another layer of indentation. So:

```typescript
if (attr === null) {
  continue;
}
```

rictic (commented):
Nice! A few requested changes that I think are all actually my fault, but that we should probably do while we're in this file

stramel:
This could probably also handle both `name` and `is` as the same value.

rictic:
Could, but code that does that is pretty vanishingly rare I think

src/collections.ts
```diff
@@ -29,7 +29,7 @@ registry.register(
       'deprecated-css-custom-property-syntax',
       'deprecated-shadow-dom-selectors',
       'dom-calls-to-native',
-      'dom-module-invalid-attrs',
+      'dom-module-name-or-is',
```
rictic: We need to change the file names, as changing a rule id is a breaking user-visible change.

rictic (changes requested):

src/collections.ts
stramel: Oh, I'm sorry. I misunderstood what you were wanting. Will update that here soon.

rictic: np!

stramel:
@rictic I believe this one is ready for review again when you get a chance :)

rictic (approved):

rictic:
So it is! I don't always get a notification for a PR without new comment activity. Maybe I need to monitor a different feed.

In any case, +1 to this!

PR #163, opened 2018-02-16
New: Warn on Element extension in createElement calls

stramel:

/cc @rictic

---
This change is [Reviewable](https://reviewable.io/reviews/polymer/polymer-linter/163)


test/create-element-extension/create-element-extension.html
```diff
@@ -0,0 +1,10 @@
+<script>
+  class Foo {
+    handler(event) {
+      document.createElement('style', 'custom-style');
```
rictic, line 4: I'm not familiar with this API. Was passing in a second string argument a supported API in v0?

src/test/polymer/create-element-extension_test.ts
```diff
@@ -0,0 +1,68 @@
...
+      `
+      document.createElement('ul', {is: 'x-expanding-list'});
+      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~`,
+      `
+      this.createElement('form', 'iron-form');
+      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~`,
+    ]);
+
+    assert.deepEqual(warnings.map((w) => w.message), [
+      'Element extension via the is attribute has been deprecated.',
```
rictic: I would quite say it's been deprecated. It looks like Firefox may implement it, and Chrome likely will too. Polymer team just discourages it because Safari has indicated that they wouldn't, so it's outside the happy no-polyfill path for many authors.

`Element extension via the is attribute is not widely supported, and is not recommended`

src/polymer/create-element-extension.ts
```diff
@@ -0,0 +1,80 @@
...
+      });
+    }
+
+    return warnings;
+  }
+
+  private isExtendingElementCall(expr: babel.Expression): boolean {
+    return babel.isCallExpression(expr) &&
+        babel.isMemberExpression(expr.callee) &&
+        (babel.isIdentifier(expr.callee.object) ||
```
rictic: For specificity, I think we should require that the callee object be named `document`

rictic (commented):
Sorry for the slow review!

src/test/polymer/create-element-extension_test.ts
stramel: Also, it was supported for a bit as a string but object seems to be the route they are looking to go.

src/polymer/create-element-extension.ts
stramel: so `document` or `this`?

test/create-element-extension/create-element-extension.html
stramel: I believe Polymer implemented it. https://github.com/w3c/webcomponents/issues/544

rictic: Arg, I see, gotcha

src/test/polymer/create-element-extension_test.ts
rictic: I see. Yeah, passing a string is pretty strongly deprecated then. Passing in an object is more of a "you probably don't want to do this"

src/polymer/create-element-extension.ts
rictic: When would we see `this.createElement`?

stramel: It wouldn't... I was mistaken. Sorry!

src/polymer/create-element-extension.ts
```diff
@@ -0,0 +1,85 @@
...
+    for (const doc of docs) {
+      babelTraverse(doc.parsedDocument.ast, {
+        noScope: true,
+        CallExpression: (path) => {
+          if (!this.isExtendingElementCall(path.node)) {
+            return;
+          }
+
+          let message: string;
+          if (babel.isStringLiteral(path.node.arguments[1])) {
```
stramel: @rictic Let me know what you think about this? I don't know quite how you would want to handle different messages for string vs object or should I just use a single message?

stramel: I'm also curious if I should only highlight the second param or the way I have with the entire call expression?

rictic: I think a separate message and a separate code is called for here. Some teams may want to ignore one warning but not the other.

rictic: Highlighting the second param is more specific and helpful I think.

stramel:
@rictic Separated the warnings and highlighted only the second argument. 

I'm not satisfied with the warning messages or codes. Let me know what you would like changed on those.

src/polymer/create-element-extension.ts
```diff
@@ -57,15 +57,18 @@ class CreateElementExtension extends Rule {
               parsedDocument: document.parsedDocument,
               code: 'create-element-string-extension',
               severity: Severity.WARNING, sourceRange,
-              message: 'Element extension via the is string is deprecated.'
+              message: stripWhitespace(`
+                Element extension via the is string is deprecated.
```
rictic: What about:

`document.createElement with a custom element tagname as the second parameter is deprecated.`

src/polymer/create-element-extension.ts
```diff
@@ -57,15 +57,18 @@ class CreateElementExtension extends Rule {
...
             }));
           } else {
             warnings.push(new Warning({
               parsedDocument: document.parsedDocument,
               code: 'create-element-is-property-extension',
               severity: Severity.WARNING, sourceRange,
-              message:
-                  'Element extension via the is property is not widely supported, and is not recommended.'
+              message: stripWhitespace(`
+                Element extension via the is property is not widely supported, and is not recommended.
```
rictic: WDYT about:

`Creating a custom element extension of a builtin element is not widely supported, and is not recommended.`

rictic (approved):
LGTM, just need to update the warning messages in the test assertions and this is ready to merge!

stramel:
Doh! Will update in a second!

rictic:
Woot!

PR #167, opened 2018-03-02
New: Validate element names

### LINKED ISSUE #95: Validate custom element tagnames
Could use something like https://www.npmjs.com/package/is-potential-custom-element-name (which is just a wrapper around a regex from the spec)

stramel:

Fixes #95 

/cc @rictic 

---
This change is [Reviewable](https://reviewable.io/reviews/polymer/polymer-linter/167)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 - Make `dom-module-invalid-attrs` rule fixable for cases where only `is` or `name` exist, and both if they use the same values.
 - `create-element-extension`: Warns when using the second parameter of `createElement` for element extension using the `is` attribute.
+- `invalid-element-name`: Warns when using an invalid element name.
+- `potential-element-naming-issue`: Warns when there are potential issues or conflicts with the element name.
 
 ## [3.0.0-pre.2] - 2018-02-16
 ### New Lint Rules
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 - Make `dom-module-invalid-attrs` rule fixable for cases where only `is` or `name` exist, and both if they use the same values.
 - `create-element-extension`: Warns when using the second parameter of `createElement` for element extension using the `is` attribute.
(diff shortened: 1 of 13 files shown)


stramel:
I'm on the fence if this should be under the Polymer or HTML folder. The naming conventions are HTML spec driven for Custom Elements and I am checking the tag name of the element. However, we are fetching "polymer-element" and highlighting the "is" value.

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 - Make `dom-module-invalid-attrs` rule fixable for cases where only `is` or `name` exist, and both if they use the same values.
 - `create-element-extension`: Warns when using the second parameter of `createElement` for element extension using the `is` attribute.
+- `invalid-element-name`: Warns when using an invalid element name.
+- `potential-element-naming-issue`: Warns when there are potential issues or conflicts with the element name.
```
rictic: We generally only list lint passes, not warnings here

test/validate-element-name/validate-element-name.html
```diff
@@ -0,0 +1,111 @@
...
+    class MyApp extends HTMLElement {
+      static get is() {
+        return 'my-app';
+      }
+    }
+    customElements.define(MyApp.is, MyApp);
+  </script>
+</dom-module>
+<!-- Error: Uppercase ASCII -->
+<dom-module id="testApp">
```
rictic: let's make this `test-App`, so that the element name is otherwise correct

src/html/validate-element-name.ts
```diff
@@ -0,0 +1,102 @@
...
+        continue;
+      }
+
+      const validationResult = validate(el.tagName);
+      if (validationResult.isValid && validationResult.message === undefined) {
+        continue;  // Valid element
+      }
+
+      const isP2 = babel.isClassDeclaration(el.astNode);
+      let sourceRange: SourceRange|undefined;
```
rictic: let's have the entire element declaration as a fallback source range, in case we don't get a source range for the tagname itself (e.g. in the case of a totally vanilla custom element, or one based on a non-polymer library).

(btw, I really love the work you've done here to get a super specific source range, and that you only do it in the rare cases where there is a problem)

stramel: Thanks 😄

rictic:
LGTM!

Issue #142, opened 2017-12-11
Call out v0 to v1 changes


stramel:
It would be nice to have linter warn and be able to fix some of the easy v0 to v1 spec changes with some Polymer changes that would allow Polymer v1 (P1) components to work with Polymer v2 (P2) legacy. These changes should allow P1 components to work with P2 without going into the full hybrid setup.

### CSS

- [x] `@apply(--foo);` to `@apply --foo;`
- [x] `var(--a, --b)` to `var(--a, var(--b))`
- [x] `:root` to `html` or `:root` to `:host > *` depending on location (shadow root vs main document) ~(#168)~
- [x] `::content` to `::slotted` ~(#160)~
- [x] `::shadow` and `/deep/` removed ~(#6) (#158)~

### HTML

- [x] `<content>` to `<slot>` (Currently listed under Polymer)
    - [x] Warns
    - [x] Fixable
- [x] `<style>` tags moved inside `<template>`

### Polymer

- [x] `document.createElement('style', 'custom-style')` not allowed anymore ~(#163)~
- [x] `is="custom-style"` to be wrapped with `<custom-style>` if not within `<template>` ~(#155)~ ~(#166)~
- [x] Check that `dom-module` has `id` attribute instead of `is` or `name`
    - [x] Warns
    - [x] Fixable ~(#162)~

Possible Additional ones
Ensure semi-colon on CSS Mixin declarations
`Polymer.dom` returns `NodeList` instead of `Array` (**NOTE:** May be difficult to detect)
`:host-context([dir=rtl]))` to `:dir(rtl)` (**NOTE:** Only available in Firefox)
`lazyRegister` no longer exists (**NOTE:** Not sure this should be a lint rule)
`this.customStyle['--my-toolbar-color'] = 'blue';` to `this.updateStyles({'--my-toolbar-color': 'blue'});`

stramel:
# Upgrade Cheatsheet
This is a simplified version of https://www.polymer-project.org/2.0/docs/upgrade. It also lists what warnings and fixes are available for each upgrade task.

## Shadow DOM template and styling

### DOM Template

- `<dom-module>` using `is` or `name` should be replaced to use `id`
	- Warns
	- Fixable: Except `dom-module`s that contain both `name` and `is`
- Move `<style>` tags inside the `<template>`
	- Warns
	- Fixable
- Update your element's DOM template to use the new `<slot>` element instead of `<content>`
	- Warns
[...]

PR #168, opened 2018-03-04
New: Replace :root selectors

stramel:

Relates to #142 

/cc @rictic

---
This change is [Reviewable](https://reviewable.io/reviews/polymer/polymer-linter/168)


stramel:
Looking for help in naming of rule code and warning message.

src/polymer/root-selector-to-html.ts
```diff
@@ -0,0 +1,106 @@
...
+        const start = node.range.start + match.index!;
+        const sourceRange = containingDoc.sourceRangeForShadyRange(
+            {start, end: start + match[0].length});
+
+        warnings.push(new Warning({
+          parsedDocument: document.parsedDocument,
+          code: this.code,
+          severity: Severity.WARNING, sourceRange,
+          message: stripWhitespace(`
+            Root should no longer be used
```
rictic: `Root` -> `The :root selector`

src/polymer/root-selector-to-html.ts
```diff
@@ -0,0 +1,106 @@
...
+          `),
+          fix: [{range: sourceRange, replacementText}]
```
rictic: We need to replace just the `:root` part, so I think we'll want to get the replacementText via a regex replace of `:root` with the passed in replacementText

src/polymer/root-selector-to-html.ts
```diff
@@ -0,0 +1,106 @@
...
+  description = stripIndentation(`
+      Warns when using :root inside an element's template or custom-style
+  `);
+
+  async check(document: Document) {
+    return [
+      ...this.generateWarnings(
+          document, isCustomStyle, 'html'),  // Check custom styles
+      ...this.generateWarnings(
+          document, isElementStyle, ':host > *'),  // Check element styles
```
rictic: If it's a dom-module without an element then it might be a style which is intended for inclusion with include, in which case we want to convert it to _both_ html and :host. See, e.g. https://github.com/PolymerElements/paper-styles/blob/master/element-styles/paper-material-styles.html#L37

If there's other parts of the selector (does that make sense in this context? I'm not sure), we also need to preserve them, so that `:root foo` becomes `:host > * foo, html foo`

src/polymer/root-selector-to-html.ts
stramel: 😆 Yeah... I really had no idea what to make this message

src/polymer/root-selector-to-html.ts
stramel: I'm not sure what you mean on this

src/polymer/root-selector-to-html.ts
stramel: hmmmm good call on the `dom-module` part, I will see if I can get that fixed as well.

As far as the addition parts to the selector, I'm not sure if it makes sense or not. I have only ever seen it in the context of `:root {` https://www.polymer-project.org/2.0/docs/upgrade#replace-root-selectors

src/polymer/root-selector-to-html.ts
```diff
@@ -0,0 +1,106 @@
...
+        predicate,
+        [],
+        dom5.childNodesIncludeTemplate);
+    if (styleTags.length === 0) {
+      return warnings;
+    }
+
+    for (const style of styleTags) {
+      const sourceRange =
+          document.parsedDocument.sourceRangeForNode(style.childNodes[0]);
```
stramel: @rictic Here, I'm getting `[ts] Object is possibly 'undefined'.` Any thoughts on how to resolve this?

src/polymer/root-selector-to-html.ts
rictic: How does this handle a selector like `:root foo`?

I think it would turn it into either `html` or `:host > *` but (I think) we want it to emit `html foo` or `:host > * foo`

src/polymer/root-selector-to-html.ts
rictic: In that case, let's not fix except in the very simple case of just `:root`

stramel:
Working on handling style modules

stramel:
Working on handling style modules

stramel:
@rictic This should be all good for review again

src/polymer/root-selector-to-html.ts
```diff
@@ -123,14 +123,20 @@ class RootSelectorToHtml extends Rule {
         const sourceRange = containingDoc.sourceRangeForShadyRange(
             {start, end: start + match[0].length});
 
+        // Only fix plain `:root` selectors
+        let fix: Replacement[] = [];
```
rictic: I think this should be instead:

```typescript
let fix: undefined|Replacement[] = undefined;
if (node.selector.match(/^:root$/) !== null) {
  fix = [{range: sourceRange, replacementText}];
}
```

Otherwise I think we tell the user that the warning could be fixed / was fixed even when we can't.

rictic:
Arg github, don't hide that comment…

src/polymer/root-selector-to-html.ts
stramel: Okay, for some reason, I thought I saw something like this elsewhere. I will make that change.

stramel:
-.- Fixing

rictic (approved):

rictic:
Great work!

PR #169, opened 2018-03-05
Fix: Don't apply fix for selector after ::slotted selector

stramel:

/cc @rictic 

I was previously not checking a case where a selector would come after the `::slotted` selector, which we can not fix safely.

---
This change is [Reviewable](https://reviewable.io/reviews/polymer/polymer-linter/169)


rictic (approved):
+1, good catch

PR #173, opened 2018-03-28
Update analyzer

rictic:
- [x] CHANGELOG.md not updated, internal change.

---
This change is [Reviewable](https://reviewable.io/reviews/polymer/polymer-linter/173)


test/call-super-in-callbacks/call-super-in-callbacks.html
```diff
@@ -80,7 +80,7 @@
    * @mixes MixinConnected
    * @polymerElement
    */
-  class GoodMixin1 extends MixinConnected(HTMLElement) {
+  class GoodMixin2 extends MixinConnected(HTMLElement) {
```
aomarks, line 83: why?

aomarks (approved):

test/call-super-in-callbacks/call-super-in-callbacks.html
rictic: Babel's scope analysis crashes when there are multiple classes with the same name in the same scope(this is illegal)

